PR #19, opened 2017-01-31
Change output paths

warpech:
Change output paths to the ones recommended in https://starcounter.gitbooks.io/guidelines/content/project_structure.html

This unifies the output path when built from VS and build.bat.

@diliger pls review


warpech:
It is not. I don't really understand `msbuild.exe` determines the output path because it seems to ignore what is in `csproj`. Perhaps some flag is missing? Anyway it now builds to the desired location so I will merge.

Issue #20, opened 2017-01-31
Clicking "Sign out" forces the DefaultTemplate


warpech:
### Steps to reproduce

1. Start Website branch `develop`
2. Start Content branch `develop`
3. Start SignIn branch `develop`
4. Go to `http://localhost:8080/signin/generateadminuser` to generate the admin user
5. Go to `http://localhost:8080/content/dynamic/apps`
6. Sign in as admin user
7. Sign out

### Expected result

You should go back to the initial page

### Actual result

You are switched to a empty page with a default template

### Screenshot

![bug](https://cloud.githubusercontent.com/assets/566463/22486041/2f2c6dea-e808-11e6-806b-01c1288723ea.gif)

warpech:
I know why this happens.

"Sign out" button in the sign in app uses a flag `HandlerOptions.SkipRequestFilters`. Website app uses a **response** filter though, and this flag does not skip that one.

In result, Website app wraps a request that should not be wrapped.

The solution requires a custom HTTP header applied in the **request** middleware that will force skipping of the **response** middleware.

This is the same fix as https://github.com/StarcounterPrefabs/Launcher/commit/39da52caab8b6f9275d05d97d0e4cac6794d71d0

warpech:
Closing as fixed, but still requires a proper fix in https://github.com/Starcounter/Starcounter/issues/4073

PR #21, opened 2017-01-31
Do not wrap requests that skip the request filter

warpech:
Fixes issue #20.

"Sign out" button in the sign in app uses a flag `HandlerOptions.SkipRequestFilters`. Website app uses a **response** filter though, and this flag does not skip that one.

In result, Website app wraps a request that should not be wrapped.

The solution requires a custom HTTP header applied in the **request** middleware that will force skipping of the **response** middleware.

This is the same fix as https://github.com/StarcounterPrefabs/Launcher/commit/39da52caab8b6f9275d05d97d0e4cac6794d71d0

I consider this to be a temporary solution until there is a flag `SkipMiddleware`, which skips both the request and the response filter.

### DIFF
--- a/src/Website/Api/ContentHandlers.cs
+++ b/src/Website/Api/ContentHandlers.cs
@@ -71,7 +71,19 @@ namespace Website {
             return false; //it's a partial
         }
 
-        protected void RegisterFilter() {
+        protected void RegisterFilter()
+        {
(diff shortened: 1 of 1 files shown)


warpech:
@chrhol, @diliger can you pls review?

warpech:
Closing as done, thanks!

PR #22, opened 2017-01-31
Fix renewing the request

### LINKED ISSUE #17: Changing sub-pages of an app reloads the whole app
### Steps to reproduce

1. Start Website branch `develop`
2. Start People branch `fix-imported-template-26`
3. Start SignIn branch `develop`
4. Go to `http://localhost:8080/signin/generateadminuser` to generate the admin user
5. Go to `http://localhost:8080/people`
6. Sign in as admin user
7. Click on Organisations in the top menu
8. Click on People in the top menu

### Expected result

As you repeat steps 7-8 in a loop, the top menu should not reload. Only the nested `CurrentPage` should reload.

### Actual result

The whole People app reloads, including the top menu as well as the nested `CurrentPage`.

You can notice the problem because the menu text `Home - People - Organizations` changes the font and size after each click.

@chrhol any idea why this happens?

### Screenshot

bug

warpech:
Fixes #17 

The problem in the code was that in the wrapped request, we did Self.GET to the `req.Uri` even though the response was already available in the `res` object.

This PR uses the actual response that came as the middleware parameter whenever possible.

@diliger pls review.

### DIFF
--- a/src/Website/Api/ContentHandlers.cs
+++ b/src/Website/Api/ContentHandlers.cs
@@ -83,7 +83,6 @@ namespace Website {
                     return response;
                 }
 
-                string[] parts = request.Uri.Split(new char[] { '/' });
                 WebUrl webUrl = Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url = ?", request.Uri).First;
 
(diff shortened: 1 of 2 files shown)


warpech:
Thanks, merging!

Issue #25, opened 2017-02-08
Remove the "Key" column from Website CMS


warpech:
Right now, all tables in Website CMS contain a "Key" column, which is totally unneeded. Please remove it.

diliger:
PR #28

diliger:
Merged PR #28 
Closing.

PR #28, opened 2017-02-09
Removed key column

diliger:
#25


diliger:
hm, for now it is not used in `javascript`

warpech:
Do you mean it *is* used? I see, then let's keep it. Please merge to `develop`, thanks!

diliger:
no, no usage in javascript.

warpech:
I can see it used in Polymer bindings in `<option>` tags, so I think it is used.

Issue #29, opened 2017-02-09
Make Website CMS appear in its own default template


warpech:
Right now, Website CMS bypasses it's own wrapping rules. 

Let's make Website CMS wrap itself in a template, meaning: migrate `StandalonePage.html` a `Templates/WebsiteCMSTemplate.html` and set up `/website/cms/*` links to be wrapped by it.

warpech:
I will work on it this weekend.

warpech:
Implemented in PR #32

warpech:
Closing as done

diliger:
PR #38

Issue #15, opened 2017-01-20
Rename some features


warpech:
Working on this app, I figured out that it is quite hard to grasp what this app does.

Perhaps renaming some features would help:

1. Rename **Templates** tab to **Surfaces**
   - Prism calls it "shells": https://msdn.microsoft.com/en-us/library/ff921098(v=pandp.40).aspx
   - Progressive Web Apps call it "shells": https://developers.google.com/web/fundamentals/architecture/app-shell
   - Material Design calls it "structures": https://material.io/guidelines/layout/structure.html#structure-ui-regions
2. In Surfaces list, rename the **Html** column to **View URI**
3. Rename **Sections** tab to **Blending points**.
   - Material Design calls it "regions": https://material.io/guidelines/layout/structure.html#structure-ui-regions
   - Prism calls it "regions": https://msdn.microsoft.com/en-us/library/ff921098(v=pandp.40).aspx
   - Marionette calls it "regions": http://marionettejs.com/docs/marionette.view.html#laying-out-views-regions
4. Rename **Urls** tab to **Catching rules**
5. In Shell rules list, rename the **Url** column to **Catch URI**
6. Rename **Maps** tab to **Pinning rules**
7. In the Pinning rules list, rename the **Section** column to **Blending point**
8. In the Pinning rules list, rename the **Url** column to **Catch URI**
9. In the Pinning rules list, rename the **Foreign url** column to **Pin URI**

Comments welcome.

diliger:
Rename just titles or all references inside the project as well?

warpech:
Whatever makes bigger sense.

Big first of all I want to get comments if you agree that this a change for better.

cc @miyconst WDYT?

warpech:
I was discussing this with @tomalec on Friday. 

I had an idea to rename **Templates** to **Surfaces*. And then to rename **Sections** to **Insertion points** or **Includes**.

@tomalec have you slept on it, do you have some new proposals here?

tomalec:
I like Surfaces.

I'd like to do another round of brainstorming, once the app's code is cleaned up, so we could reduce even more :>

warpech:
I have changed the first post to contain the current proposal.

Please change texts, urls, view-model names and file names. Do NOT change Simplified class names.

This is not final change in the UI, in a next issue we want to streamline it more to look like CloudFlare:

[...]

PR #32, opened 2017-02-13
Make Website CMS be wrapped in DefaultTemplate

warpech:
Implements #29 

This required to split the app in two: `Website` and `WebsiteProvider`

As a side feature, WebsiteProvider allows to nest templates in other surfaces, which is demonstrated by adding a Help panel on Website CMS Templates page.

To test, make sure you run `http://localhost:8080/website/resetdata` to load additional maps (SidebarTemplate) that have been created.

@diliger please review.

Screenshot:

2017-02-13 11_54_58-website - microsoft visual studio


diliger:
Well, i read it and run it, it works.

But i want to get better into the logic.
Now there are two apps, Website and WebsiteProvider
From description: 

> Website app gives "Website CMS"
> WebsiteProvider app gives section wrapping

What happens inside, what handlers affected
for example if I request the URI `http://127.0.0.1:8080/website/cms/templates`

I see a handler in Website app 
`Handle.GET("/website/cms/templates", () => {`
And how the WebsiteProvider affects the output result?

diliger:
Why the default `/Website` URI handles **WebsiteProvider** not the **Website** app?

warpech:
Technically, the apps needed to be split, because I wanted to wrap Website CMS in a template. This was not possible in the same app, because it would require merging two responses from the same app as one level in JSON (currently we support only one response at a single level).

Conceptually, splitting into two apps has more benefits than that:

- Website CMS app is used for configuration of templates/surfaces and can be stopped when you are done with configuring
- WebsiteProvider is lightweight and used for providing surfaces.

If the above explanation is enough, I will add it to README.md. Glad that you've asked.

> Why the default `/Website URI` handles **WebsiteProvider** not the **Website** app?

That's clearly a mistake, thanks for reporting! I will fix.

warpech:
> I see a handler in Website app 
`Handle.GET("/website/cms/templates", () => {`
And how the WebsiteProvider affects the output result?

This handler provides a partial for configuring templates/surfaces. WebsiteProvider wraps the response from this handler (or any other handler in any other app) and puts that response in a section/blending point of a template/surface.

warpech:
I have added the above explanation in README and renamed handlers in WebsiteProvider to /WebsiteProvider/

diliger:
Well, seems i got it, where the magic does.

> `Application.Current.Use((Request request, Response response) => {`

this executes for every request, looks for the template based on the URI in the database and wrap the response into template. 

Just for me the name WebsiteProvider is not very clear..
Maybe WebsiteTemplates or TemplatesProvider.

And why **Default** template means **Final**?

> `master.IsFinal = template.Default;`

diliger:
Fixed code style for better look in ContentHandlers

warpech:
Yes, it is exactly as you described.

> Just for me the name WebsiteProvider is not very clear..
Maybe WebsiteTemplates or TemplatesProvider.

"Website" and "WebsiteProvider" is not very bad, either. I think that "Template" and "TemplateProvider" would be even less clear.

The issue #15 renames the term "Templates" to "Surfaces".  So maybe SurfaceDesigner and SurfaceProvider would be better?

diliger:
Okay, let it be "Website" and "WebsiteProvider".

warpech:
> And why **Default** template means **Final**?
> `master.IsFinal = template.Default;`

WebsiteProvider works recursively, allowing you to wrap a surface in another surface.

Because of that, it would go into an infinite look of wrapping default template in the default template. The `IsFinal` property prevents that.

warpech:
Thanks for the review, merging.

Issue #24, opened 2017-02-08
open Website default page stops all applications


diliger:
when I click the button Open for **Website** app in **Starcounter Administrator** 
page for `localhost.../Website` opens, results to `URI not found` and all started applications die.

is that appropriate behaviour? 
![image](https://cloud.githubusercontent.com/assets/15857369/22742475/e853530e-ee27-11e6-9517-2595e72f70af.png)

warpech:
@diliger thanks for reporting! It is certainly not appropriate.

When reporting bugs, please provide information like:

- Starcounter version
- App version
- Error from the error log

---

I **can** reproduce it with Website app (`develop` branch) and Starcounter 2.3.0.5512

In Administrator Log I can see an exception:

```
Non database type Starcounter.Metadata.MetadataEntity of property Type in class Starcounter.Internal.Metadata.SetSpecifier
```

@andwah, @per-samuelsson any ideas what is happening?

andwah:
Sorry, no ideas just by reading this issue.

per-samuelsson:
> In Administrator Log I can see an exception:
>> Non database type Starcounter.Metadata.MetadataEntity of property Type in class Starcounter.Internal.Metadata.SetSpecifier

As an *exception*, really? Otherwise I would have suggested that thing was not related to the crash of the host, because it's already a [known issue](https://github.com/Starcounter/Starcounter/issues/3937). (But apparently by design)

warpech:
@per-samuelsson you a right, this is an unrelated warning...

So yes, this is a crash that is caused by the handler `/Website` in the Website app. It happens also when you visit the URI directly, not from Administrator.

Visual Studio shows me that calling the handler throws:

```
System.StackOverflowException was unhandled
Message: An unhandled exception of type 'System.StackOverflowException' occurred in Starcounter.XSON.dll
```

The above exception is not registered in the error log.

After that, all running applications stop.

@diliger can you investigate what line causes the StackOverflowException?

chrhol:
I will take a look as well. Sounds like some internal problem in Starcounter.XSON or that a sibling (=merged json) is attached in a way that causes an infinite loop.

I have a point to add checks for this inside TypedJSON to avoid stackoverflow, but haven't gotten around to implement it yet. I will raise the priority on it though.

chrhol:
Something is messed up with the merged json/siblings

One instance of json in website is merged with a parent of itself.

```
x = {Website.LayoutPage}
[...]

PR #38, opened 2017-02-17
Handle /website #24 

diliger:
#24 
Added Handle for GET(/Website) and
AboutPage for GET(/WebsiteProvider)

### DIFF
--- a/src/WebsiteProvider/Api/ContentHandlers.cs
+++ b/src/WebsiteProvider/Api/ContentHandlers.cs
@@ -39,11 +39,11 @@ namespace WebsiteProvider
 
             Handle.GET("/WebsiteProvider", () =>
             {
-                WrapperPage master = GetLayoutPage();
+                //WrapperPage master = GetLayoutPage();
 
-                master.WebTemplatePage.Data = null;
+                //master.WebTemplatePage.Data = null;
 
-                return master;
+                return new AboutPage();
             });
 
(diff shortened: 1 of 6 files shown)


diliger:
cc @warpech

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -39,11 +39,11 @@ public void Register()
...
             {
-                WrapperPage master = GetLayoutPage();
+                //WrapperPage master = GetLayoutPage();
 
-                master.WebTemplatePage.Data = null;
+                //master.WebTemplatePage.Data = null;
 
-                return master;
+                return new AboutPage();
             });
```
warpech: Please remove the comments, the code will not be needed at all.

`new AboutPage()` cannot be rendered right now, because WebsiteProvider wraps the page in a Json.

To fix this, I have no better idea than remove AboutPage and return a simple string, similarly how it is done in SignIn: https://github.com/StarcounterPrefabs/SignIn/blob/a750754e9c2af6502b78d717096850ccbcef363d/src/SignIn/Api/MainHandlers.cs#L137

Text responses are not wrapped in surfaces. Only `Json` responses are wrapped.

diliger:
Removed AboutPage

warpech:
Looking good! Please merge

Issue #31, opened 2017-02-11
Add a button to Restore the default surfaces


warpech:
We should make it impossible to delete the default template, because that is hard to recover. 

Or allow to revoke the default template with some button.

warpech:
How about fixing it differently?

1) Clicking on an unchecked "Default", unchecks the previous default and checks the new one
2) Clicking on an checked "Default" does nothing
3) It is not possible to delete a template marked as "Default"

The same should apply for Sections:

1) When you create a Template/Surface, it always has one "Main" section/blending point that is marked as Default.
1) Clicking on an unchecked "Section", unchecks the previous default section in that template and checks the new one
2) Clicking on an checked "Default" does nothing
3) It is not possible to delete a section marked as "Default"

diliger:
But in that way it is possible to have only one **Default** `Surface` at a time.
And I think that is not correct.
For example we may have `LauncherTemplate` and `DefaultTemplate` both as **Default**, don't we?

warpech:
I think that it doesn't make sense to have two "Defaults", because what would that mean? This SQL query only cares about the first Default found: https://github.com/StarcounterPrefabs/Website/blob/8633a0cb821c5dbee341b18b5f1e2ba0d3bcba7a/src/WebsiteProvider/Api/ContentHandlers.cs#L103

Anyway, please disregard my above comments, because they conflict with PR #43.

diliger:
I think that means it is the Final and shouldn't be wrapped again.

diliger:
Closing as done.

PR #39, opened 2017-02-19
save changes can not be performed if no default templates in the list #31 

diliger:
added Restore button to restore default templates
#31


diliger:
But it ends up with error if press Reset button.

> `System.Exception: ScErrDataBindingForJsonException (SCERR14012): An exception occurred when a databinding for a property in json was used. Property 'Html' in 'WrapperPage.WebTemplatePage'.
Version: 2.3.0.5651.
Help page: https://github.com/Starcounter/Starcounter/wiki/SCERR14012.
   at lambda_method(Closure , Json )
`
> `Starcounter.DbException: ScErrInvalidObjectAccess (SCERR4174): The operation failed because of an invalid object access.`

How that can be fixed?

diliger:
cc @warpech

warpech:
@chrhol in https://github.com/StarcounterPrefabs/Website/pull/39#issuecomment-280918827, @diliger has a problem because one app has a `Json` bound to a data object that is deleted in another app. 

What can we do about it? I would like to have some callback when data binding fails and assign some other ("default") object to it, when the original object is missing. Is that possible?

warpech:
@diliger, the problem that you describe in https://github.com/StarcounterPrefabs/Website/pull/39#issuecomment-280918827 is caused by https://github.com/Starcounter/Starcounter/issues/2140 and solving it will take time.

So we need to finish this PR knowing that there is this limitation. When you press "Restore", you need to reload the page.

Can you please continue with my proposal from https://github.com/StarcounterPrefabs/Website/issues/31#issuecomment-281316641, if you agree with it?

src/Website/wwwroot/Website/viewmodels/CmsSurfacesPage.html
```diff
@@ -48,4 +52,16 @@
...
         </fieldset>
     </template>
+    <script>
+        (function () {
+            var template = (document._currentScript || document.currentScript).previousElementSibling;
+
+            template.restore = function () {
+                template.set("model.Restore$", template.model.Restore$ + 1);
+                setTimeout(function () {
+                    window.location = window.location.href;
```
warpech: Please change to [`window.location.reload(true);`](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload) and add a comment why this is needed.

src/Website/wwwroot/Website/viewmodels/CmsSurfacesPage.html
```diff
@@ -3,6 +3,7 @@
         <fieldset>
             <legend>
                 <span class="pull-right">
+                    <span class="text-danger">{{model.Error}}</span>
```
warpech: Please remove this, because it conflicts with PR #43.

src/Website/wwwroot/Website/viewmodels/CmsSurfacesPage.html
```diff
@@ -15,6 +16,9 @@
                             onmousedown="++this.value;" title="Create template">
                         Create
                     </button>
+                    <button type="button" class="btn btn-xs btn-default" on-click="restore" title="Restore default">
+                        Restore
```
warpech: Please change the button text to "Restore default surfaces"

diliger:
Pushed.

src/Website/ViewModels/CmsSurfacesPage.json.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
```
warpech: `using System.Linq;` can be removed now.

diliger: Actually I dislike the idea to remove default usings.
They dont eat any resources but every time to add them and remove back is annoying...
But that is IMHO

Issue #33, opened 2017-02-13
DefaultTemplate TopBar appears wrong


warpech:
### Steps to reproduce

1. Start Website and WebsiteProvider (branch `feature/29-cms-default-template`)
2. Start SignIn branch `develop`
2. Start CompositionEditor branch `develop`
2. Call `http://localhost:8080/website/resetdata` to load default mappings
4. Go to `http://localhost:8080/signin/generateadminuser` to generate the admin user
5. Go to `http://localhost:8080/website/cms`
6. Sign in as admin user

### Expected result

In the top bar, on the right side of the screen, there should be two icons in the same row.

### Actual result

In Chrome, the icons appear in two different rows.

In Firefox, the icons appear on the left side on the screen and in two different rows.

This should be fixed the similarly to how it works in Launcher: https://github.com/StarcounterPrefabs/Launcher/blob/develop/src/Launcher/wwwroot/Launcher/viewmodels/LauncherUser.html, but without calling `:host` from the child page. Rather than that, fix the styles in: https://github.com/StarcounterPrefabs/Website/blob/feature/29-cms-default-template/src/Website/wwwroot/Website/templates/DefaultTemplate.html

@diliger can you pls take a look?
 
##### Browsers affected

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

#### Chrome

![2017-02-13 11_54_58-website - microsoft visual studio](https://cloud.githubusercontent.com/assets/566463/22881200/ca3d1302-f1e5-11e6-92e4-e8a9e442693a.png)

#### Firefox

![2017-02-13 11_57_59-website - microsoft visual studio](https://cloud.githubusercontent.com/assets/566463/22881207/cea3cf58-f1e5-11e6-9da1-67f86f76851c.png)

diliger:
after some investigations.. I have a question here
to display it fine with css i should apply `display: flex;` for very nested `juicy-composition`
see the picture
![image](https://cloud.githubusercontent.com/assets/15857369/23097583/1b83bf66-f648-11e6-8495-9c88c47099cb.png)

the style created in `DefaultTemplate` is

> `.website-defaulttemplate-topbar ::content > starcounter-include`

but in my case it should be 

> `.website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition > starcounter-include > juicy-composition`

I think it is a bit too long and unstable.

[...]

PR #40, opened 2017-02-21
css fix for TopBar #33

diliger:
Css rule works fine in Chrome but absolutely ignored by FireFox.
Seems it doesn't understand the ::content selector.
How that may be fixed?
#33

### DIFF
--- a/src/Website/wwwroot/Website/templates/DefaultTemplate.html
+++ b/src/Website/wwwroot/Website/templates/DefaultTemplate.html
@@ -16,31 +16,63 @@
                 color: #FFF;
                 padding: 10px;
             }
-                /** for Shadow DOM v0 compilant browsers */
-                .website-defaulttemplate-topbar ::content > starcounter-include,
-                .website-defaulttemplate-topbar ::content starcounter-include > juicy-composition {
-                    display: flex;
+            /* Blending points and their elements distributed in topbar surface */
+            /* For browsers without CSS Grid Layout and witout @supports make topbar a flexbox*/
+            /** for Shadow DOM v0 compilant browsers */
+            .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition,
+            .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition > starcounter-include > juicy-composition {
+                display: flex;
+                align-items: center;
+                justify-content: flex-end;
+            }
+            /** for polyfilled browsers */
+            .website-defaulttemplate-topbar > starcounter-include > juicy-composition,
+            .website-defaulttemplate-topbar > starcounter-include > juicy-composition > starcounter-include > juicy-composition {
+                display: flex;
+                align-items: center;
+                justify-content: flex-end;
+            }
+            /* For ones with CSS Grid us it with 5px gutter */
+            @supports( (display: grid) and (grid-column-gap: 5px) ) {
+                /** For browsers that supports Shadow DOM v0 */
+                .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition,
+                .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition > starcounter-include > juicy-composition {
+                    display: grid;
                     align-items: center;
-                    justify-content: flex-end;
-                    margin-right: 5px;
-                }
-                /** for polyfilled browsers */
-                .website-defaulttemplate-topbar > starcounter-include,
-                .website-defaulttemplate-topbar starcounter-include > juicy-composition {
-                    display: flex;
-                    align-items: center;
-                    justify-content: flex-end;
+                    grid-column-gap: 5px;
+                    grid-auto-flow: column;
                     margin-right: 5px;
                 }
 
-                    /** for Shadow DOM v0 compilant browsers */
-                    .website-defaulttemplate-topbar ::content > starcounter-include > starcounter-include {
-                        margin-left: 10px;
-                    }
-                    /** for polyfilled browsers */
-                    .website-defaulttemplate-topbar ::content > starcounter-include > starcounter-include {
-                        margin-left: 10px;
-                    }
+                /** duplicated rule for polyfilled browsers */
+                .website-defaulttemplate-topbar > starcounter-include > juicy-composition,
+                .website-defaulttemplate-topbar > starcounter-include > juicy-composition > starcounter-include > juicy-composition {
+                    display: grid;
+                    align-items: center;
+                    grid-column-gap: 5px;
+                    grid-auto-flow: column;
+                    margin-right: 5px;
+                }
+            }
+            /* For browsers w/o Grid break SD boundary and add gutter as margin */
+            @supports( not ((display: grid) and (grid-column-gap: 5px)) ) {
+                .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition > *,
+                .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition > starcounter-include > juicy-composition > *,
+                /*
+                 Shadow DOM v1 in-compatible Shadow DOM piercing selector
+                 styling such things may break default compositions delivered by apps
+                */
+                .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition::shadow>*,
+                .website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition > starcounter-include > juicy-composition::shadow>* {
+                    margin-right: 5px;
+                }
+                /** duplicated rule for polyfilled browsers */
+                .website-defaulttemplate-topbar > starcounter-include > juicy-composition > *,
+                .website-defaulttemplate-topbar > starcounter-include > juicy-composition > starcounter-include > juicy-composition > * {
+                    margin-right: 5px;
+                }
+            }
+
         </style>
         <section class="website-defaulttemplate-topbar">
             <content select="[slot='Website/topbar']"></content>
--- a/src/Website/wwwroot/Website/templates/DefaultTemplate.html
+++ b/src/Website/wwwroot/Website/templates/DefaultTemplate.html
@@ -16,10 +16,16 @@
                 color: #FFF;
                 padding: 10px;
             }
-
+                /** for Shadow DOM v0 compilant browsers */
                 .website-defaulttemplate-topbar ::content > starcounter-include,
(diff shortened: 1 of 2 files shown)


warpech:
I am afraid that we need help from @tomalec here.

@tomalec, the steps to reproduce are in #33. There are two problems:

1) In Chrome, there is no 5px margin between the icons
2) In Firefox, the flex model is not applied at all to the top bar

This affects both LauncherTemplate and DefaultTemplate.

tomalec:
What URL should I look for? as after starting Website and Website provider I get `URI not found: GET /website` for http://localhost:8080/website

warpech:
See step 6 in reproduction instructions

tomalec:
> 1. In Chrome, there is no 5px margin between the icons

Could you tell me where is `5px` defined for icons? I can only see `5px` as a `margin-right` for entire row https://github.com/StarcounterPrefabs/Website/blob/8e96d17dc930eae461416711e365744f7cbcaf0d/src/Website/wwwroot/Website/templates/DefaultTemplate.html#L27

warpech:
Ah you are right, it is only in LauncherTemplate: https://github.com/StarcounterPrefabs/Website/blob/8e96d17dc930eae461416711e365744f7cbcaf0d/src/Website/wwwroot/Website/templates/LauncherTemplate.html#L185-L188

DefaultTemplate was expected to copy the working solution from LauncherTemplate. @diliger can you please copy the 5px margin between the icons from LauncherTemplate to DefaultTemplate?

warpech:
Ah you are right, it is only in LauncherTemplate: https://github.com/StarcounterPrefabs/Website/blob/8e96d17dc930eae461416711e365744f7cbcaf0d/src/Website/wwwroot/Website/templates/LauncherTemplate.html#L185-L188

But the problem 1. from https://github.com/StarcounterPrefabs/Website/pull/40#issuecomment-281346382 persists. The 5px right margin between icons doesn't work in Chrome. It seems that the `> *` part of the selector to applies to the content node, not to the Shadow DOM of the nested `juicy-composition`.

The problem 2 persists, too.

tomalec:
I'm still investigating the problem 2.

The simplest solution would be to duplicate selectors:
```css

                .website-defaulttemplate-topbar > starcounter-include,
                .website-defaulttemplate-topbar starcounter-include > juicy-composition
                 {
                    display: flex;
                    align-items: center;
                    justify-content: flex-end;
                    margin-right: 5px;
                }
                /** for Shadow DOM compilant browsers */
                .website-defaulttemplate-topbar ::content > starcounter-include,
                .website-defaulttemplate-topbar ::content starcounter-include > juicy-composition {
```

I'm now investigating why `/* @polyfill selector */` does not work

tomalec:
Speaking of `juicy-composition > *`
Naturally it selects all (light DOM) direct children of `juicy-composition`
image

In this case `[template, starcounter-include, template, template]` Which elements you would like to select?

tomalec:
If you would like to pierce through Shadow DOM you can use deprecated `::shadow` but in V1 it will be removed https://hayato.io/2016/shadowdomv1/#shadow-piercing-combinators
I'd suggest adding that rule in Shadow Root for icons area shadow dom

warpech:
> Which elements you would like to select?

I would like to have a 5px space between icons. They can be any elements, and they can be wrapped in Shadow DOM of the toolbar view.

> I'd suggest adding that rule in Shadow Root for icons area shadow dom

Yeah, this sounds like the only way to me. But that requires additional HTML to the merged HTML, doesn't it?

tomalec:
Ok, I'll investigate further.

Is the 2. solution fine enough, can we leave it and postpone further investigation for V1?

tomalec:
For 1.

The icon elements does have margin between them. However, div with white background in Shadow DOM does not have one.
The cleanest solution would be to use a CSS Grid layout instead of flexbox, then assign `grid-gap` as a property of a container - as this is in fact what we would like to achieve: specify the way container should layout children and provide some gutter between them.

warpech:
I guess CSS `@supports` allows us to use `grid-gap` if the browser supports CSS Grid Layout, otherwise fall back to `display: flex` with no margin. @tomalec could you please implement it that way?

tomalec:
By now I would use `::shadow` then either investigate solutions deeper while migrating to V1, or wait for CSS Grid, or wait for new `declarative-shadow-dom` which should allow you to do something like
```html
<starcounter-include view-model="{{path.to.top.right.corner}}">
    <template is="declarative-shadow-dom">
        <style>:host > * { margin-right: 5px;}</style>
    </template>
</starcounter-include>

**Comment by tomalec** [comment]
I have figured, that to fix 2. in clean and platform wide level, `juicy-composition` or `starcounter-include` could run `WebComponents.ShadowCSS.shimStyling` for loaded compositions, as webcomponents.js (HTML Imports polyfill) does not call it automatically on our `<styles>`, as they are within `<template>` at that time.

But, as we agreed to stop V0 development https://github.com/Starcounter/RebelsLounge/issues/66, I'd leave it with duplicated rules.

**Comment by tomalec** [comment]
@warpech I've finished with my changes, please review.
You can check grid rendering in Chrome Canary (should be enabled by default)
(In case you would have color problems as I did https://productforums.google.com/forum/#!topic/chrome/BtX8ZGVeDLQ )

**Comment by warpech** [comment]
Thank you! `LauncherTemplate.html` now looks as desired in Chrome (Experimental Web Platform Features enabled), Chrome (EWPF disabled), Firefox, Edge and Safari.

`DefaultTemplate.html` still does not have the 5px margin. @tomalec could you please try to fix that?

---

Are you convinced that this:

```css
                .launcher-layout__top-right ::content starcounter-include > juicy-composition > *,
                /*
                 Shadow DOM v1 in-compatible Shadow DOM piercing selector
                 styling such things may break default compositions delivered by apps
                */
                .launcher-layout__top-right ::content starcounter-include>juicy-composition::shadow>* {
                    margin-right: 5px;
                }
                /** duplicated rule for polyfilled browsers */
                .launcher-layout__top-right starcounter-include > juicy-composition > * {
                    margin-right: 5px;
                }
```

Looks better than:

```css
                .launcher-layout__top-right ::content starcounter-include > juicy-composition > *,
                /*
                 Shadow DOM v1 in-compatible Shadow DOM piercing selector
                 styling such things may break default compositions delivered by apps
                */
                .launcher-layout__top-right ::content starcounter-include>juicy-composition::shadow>*,
                /** duplicated rule for polyfilled browsers */
                .launcher-layout__top-right starcounter-include > juicy-composition > * {
                    margin-right: 5px;
                }
```

For me, the latter looks more readable.

tomalec:
I'm not saying it's more readable. I do agree it's not. But simply only
first one works.
I'll chec DefaultTemplate tomorrow.

22 lut 2017 22:27 "Marcin Warpechowski" <notifications@github.com>
napisał(a):

tomalec:
I'm not saying it's more readable. I do agree it's not. But simply, it's  only
the first one which works.

And by what I was saying [here](https://github.com/StarcounterPrefabs/Website/pull/40#issuecomment-281773538), after some effor, and by cost of performance in `starcounter-include`, we could achieve slightly more readable form:
```css
/* @polyfill .launcher-layout__top-right starcounter-include > juicy-composition > * */
.launcher-layout__top-right ::content starcounter-include > juicy-composition > *,
/*
 Shadow DOM v1 in-compatible Shadow DOM piercing selector
 styling such things may break default compositions delivered by apps
*/
.launcher-layout__top-right ::content starcounter-include>juicy-composition::shadow>* {
    margin-right: 5px;
}
```

I'll check `DefaultTemplate` tomorrow.

warpech:
> I'm not saying it's more readable. I do agree it's not. But simply, it's only
the first one which works.

Sorry, I missed that. Then let's keep it as is.

warpech:
> I'll check `DefaultTemplate` tomorrow.

@tomalec were you able to take a look?

tomalec:
Done, I have also unified result of grid and flex fallback for `LauncherTemplate`.
I don't know what was `.website-defaulttemplate-topbar ::content > starcounter-include > starcounter-include` for (as there is no such element) so I removed it.

Tested in 
 - Chrome Stable 56 (Shadow DOM + Flex fallback)
 - Chrome Canary 57 (Shadow DOM + Grid)
 - Firefox Stable 51 (SD Polyfill + Flex fallback)
 - Firefox Dev 52 (SD Polyfill + Grid)
 - IE & Edge (SD Polyfill + Flex fallback)

warpech:
> Done, I have also unified result of grid and flex fallback for LauncherTemplate.
I don't know what was .website-defaulttemplate-topbar ::content > starcounter-include > starcounter-include for (as there is no such element) so I removed it.

That's ok, thanks!

There are still two problems, @tomalec could you please take a look:

1) The CSS selectors in `DefaultTemaplate.html` and `LauncherTemplate.html` should not be that much different. 

Right now Default uses:

```
.website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition,
.website-defaulttemplate-topbar ::content > starcounter-include > juicy-composition > starcounter-include > juicy-composition {
```

And Launcher uses:

```
launcher-layout__top-right ::content starcounter-include>juicy-composition
```

I think that Launcher's version, without excessive use of child selector is cleaner for the eye. Could you please copy that to DefaultTempalte.html?

2) In LauncherTemplate.html, SignIn app icon looks wrong after signing in:

2017-02-24 14_38_21-website

tomalec:
`launcher-layout__top-right ::content starcounter-include>juicy-composition`

Select in fact **any** nested view **any** deep - is that what we want? If so, I'll change it.

---

> In LauncherTemplate.html, SignIn app icon looks wrong after signing in:

What browser it is? in all mine it looks fine
image

warpech:
It turns this might be not the fault of your changes.

I can reproduce this by running Website (`defaultTemplate-topBar-fix` branch) and SignIn (`feature/Show-user-image` branch).

warpech:
> Select in fact **any** nested view **any** deep - is that what we want? If so, I'll change it.

For the topbar, yes, there should not be much nesting, so let's embrace the simpler CSS selector.

warpech:
So, the joke is on me! Actually the SignIn icon displays wrongly in `LauncherTemplate.html` because of the too simple selector: `launcher-layout__top-right ::content starcounter-include>juicy-composition`

@tomalec please use the explicit child selectors in `LauncherTemplate.html`, the way they are done in `DefaultTemplate.html`

PR #43, opened 2017-02-24
Added "Final" flag for Catching URIs and removed "Default" flag for Surfaces

un-tone:
For https://github.com/StarcounterPrefabs/Website/issues/41

### DIFF
--- a/src/Website/Helpers/DataHelper.cs
+++ b/src/Website/Helpers/DataHelper.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using Starcounter;
+﻿using Starcounter;
 using Simplified.Ring6;
 
 namespace Website
@@ -37,7 +36,6 @@ namespace Website
         {
             var surface = new WebTemplate()
             {
-                Default = true,
                 Name = "DefaultTemplate",
                 Html = "/Website/templates/DefaultTemplate.html"
             };
@@ -56,6 +54,13 @@ namespace Website
                 Default = true
             };
 
+            var catchAllUrl = new WebUrl()
+            {
+                Template = surface,
+                Url = null,
+                IsFinal = true
+            };
+
             new WebMap() { Section = topbar, ForeignUrl = "/signin/user", SortNumber = 1 };
         }
 
@@ -63,7 +68,6 @@ namespace Website
         {
             var surface = new WebTemplate()
             {
-                Default = false,
                 Name = "SidebarTemplate",
                 Html = "/Website/templates/SidebarTemplate.html"
             };
@@ -95,7 +99,6 @@ namespace Website
         {
             var surface = new WebTemplate()
             {
-                Default = false,
                 Name = "AppHubTemplate",
                 Html = "/Website/templates/AppHubTemplate.html"
             };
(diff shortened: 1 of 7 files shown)


un-tone:
Related to the PR https://github.com/StarcounterPrefabs/Simplified/pull/45

src/Website/Helpers/DataHelper.cs
```diff
@@ -56,14 +54,20 @@ public void GenerateDefaultSurface()
                 Default = true
             };
 
+            var catchAllUrl = new WebUrl()
+            {
+                Template = surface,
+                Url = null,
+                IsFinal = true
+            };
```
warpech: I think that a similar Catch All + IsFinal rule should be added to `GenerateAppHubSurface` and `GenerateLauncherSurface`

From the current catching rules, only SidebarSurface rule for `/website/cms/surfaces` is not final.

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -82,26 +81,18 @@ public void Register()
...
                 WebUrl webUrl = Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url = ?", requestUri).First;
 
                 if (webUrl == null)
                 {
                     string wildCard = GetWildCardUrl(requestUri);
 
-                    webUrl = Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url = ?", wildCard).First;
+                    webUrl = Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url = ?", wildCard).First
+                             ?? Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url IS NULL AND wu.IsFinal = ?", true).First
+                             ?? Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url IS NULL").First;
```
warpech: Will this catch a manually created catch-all rule? I think the `Url` value will be an empty string, so the last two SQLs should check for null or empty string.

warpech:
@un-tone can you also please update README.md to be relevant after this change?

un-tone:
@warpech could you clarify what info do you want to be added to the README.md exactly? I'm not confident since it contains only publishing instructions currently.

src/Website/Helpers/DataHelper.cs
un-tone: Set IsFinal flag for default ApHubTemplate catching rules

src/WebsiteProvider/Api/ContentHandlers.cs
un-tone: Fixed the conditions.

warpech:
I mean README.md in the `develop`. You will know when you see it, just search for the word "Default".

Anyway. I will merge `develop` to `master` now to avoid confusion.

warpech:
Perfect explanation in README.md, thanks

PR #57, opened 2017-03-08
Throw an exception when an application doesn't provide "Html"

un-tone:


warpech:
If some app exposes a REST API, then in fact you would want to return a JSON response without `Html` property. 

So I think that a desired solution to https://github.com/Starcounter/Starcounter/issues/4115#issuecomment-285066563 would be to make WebsiteProvider wrap only reponses that are of type `Json` and have `Html` property.

If your response is not `Json` or it is a `Json` without a `Html` property, WebsiteProvider should be bypassed.

WDYT @un-tone?

Issue #63, opened 2017-03-13
Incorrect merging of JSON when adding another app's data into the same Section


un-tone:
### Steps to reproduce
1. Start Website (branch `develop`), WebsiteProvider (branch `develop`), SignIn (branch `develop`)
2. Start People (branch `develop`) and KitchenSink (branch `test/External-App-Link-People`)
3. Go to `http://localhost:8080/KitchenSink`
4. Click on the "People app" link (below in the navigation area)
5. Check Console logs and Network tabs

### Expected result
Shows the Persons page of the People app

### Actual result
Throws an error `Uncaught TypeError: Cannot read property 'CurrentPage' of undefined`.
The `WebsiteProvider` should send in the response a correct `add` operation data for the People app data and a `remove` operation for KitchenSink data.

##### Involved apps' version
 - StarcounterApps/Website@04fd00046e025d618fdbf89c211b28bc54f4d0dd
 - StarcounterApps/KitchenSink@fae9156da92173ab3f70d79bed38d5dcb254d149
 
##### Browsers affected

- [x] Chrome

### Screenshot
![image](https://cloud.githubusercontent.com/assets/1625378/23853382/5a1d767e-07fd-11e7-959a-c695694e23ae.png)

warpech:
@chrhol could you pls investigate this? Looks like a server side issue with patch generation.

chrhol:
I've found a couple of bugs. One I have fixed (only locally so far though). The other one comes from reusing the instance of the sections, where patches for the new and removed siblings is not detected. This I have yet to fix. I will create an issue in the Starcounter repo for that one.

I found one more thing, I tried to change so that the section is not reused (as a workaround) but instead always sends out a replace of the whole row, by commenting out [these lines](https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteProvider/Api/ContentHandlers.cs#L208-L209)

The patches looks correct, however the viewmodel doesn't seem to be updated. For this one I have no idea why, there's nothing in the console. The patch just dont get applied.
![patch](https://cloud.githubusercontent.com/assets/1658979/24113417/0721ef72-0d9d-11e7-931a-294b9fb472be.PNG)
![model](https://cloud.githubusercontent.com/assets/1658979/24113419/0ac6a14a-0d9d-11e7-9cfc-0d55f9d75f85.PNG)

chrhol:
[...]

Issue #58, opened 2017-03-08
Inifinte loop when no Catching Rule is default


joozek78:
### Steps to reproduce

1. Untick "final" checkbox beside all Catching Rules with empty URI
1. (Try to) load any page

### Expected result

Stuff loads and works. If no catching rule matches, then the page is loaded as-is

### Actual result

An infinite loop. Debug indicates it's somewhere inside here:

```
[External Code]	
WebsiteProvider.exe!WebsiteProvider.ContentHandlers.UpdateTemplateSections(string requestUri, Starcounter.Response response, WebsiteProvider.WebTemplatePage content, Simplified.Ring6.WebUrl url) Line 201	C#
WebsiteProvider.exe!WebsiteProvider.ContentHandlers.Register.AnonymousMethod__3_2(string requestUri) Line 102	C#
[External Code]	
WebsiteProvider.exe!WebsiteProvider.ContentHandlers.RegisterFilter.AnonymousMethod__6_1(Starcounter.Request request, Starcounter.Response response) Line 138	C#
[External Code]	
```

##### Starcounter version

2.3.0.5851

##### Involved apps' version

StarcounterApps/Website@0b1fa612cb1445d27f072a874b07b573fd9f2e16

un-tone:
Created PR https://github.com/StarcounterApps/Website/pull/65.
Also I want to clarify, the application requires to have at least "final" or "catch-all" URI rule for all requests.

joozek78:
It would be nice if it didn't required that, but that's actually quite easy to work around. More importantly, the application should tell the user about its requirements, not enter an infinite loop nor throw mysterious exception.

warpech:
Once again I would start from taking inspiration from https://www.cloudflare.com/features-page-rules/must-use-page-rules/

> Only one Page Rule will trigger per URL, so it is helpful if you sort Page Rules in priority order, and make your URL patterns as specific as possible.

----

Let's translate it to our jargon:

1. Only one Catching Rule will trigger per Catch URI, so it is helpful if you sort Catching Rules in priority order, and make your URI patterns as specific as possible.

And let's add three more constrains:

2. A matched Catching Rule will cause a new run of the Catching Rules loop, using the Surface URI as the Catch URI in the new run
3. A matched Catching Rule can only be triggered once in the Catching Rules loop
4. The Catching Rules loop finishes when a matched Catching Rule is Final or if the URI does not matched any Catching Rule

[...]

PR #65, opened 2017-03-14
Fix catch-all rule handling

un-tone:
For https://github.com/StarcounterApps/Website/issues/58


src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -93,7 +93,7 @@ public void Register()
 
                 this.CurrentTemplate = template;
                 WrapperPage master = GetLayoutPage();
-                master.IsFinal = webUrl.IsFinal;
+                master.IsFinal = webUrl.IsFinal || string.IsNullOrEmpty(webUrl.Url);
```
joozek78: What if there is no rule that has empty URL?
Also I don't think this change here is good - what if someone legitimately wants a rule that catches all, but isn't final?

warpech: Added my comments in https://github.com/StarcounterApps/Website/issues/58#issuecomment-286901166

un-tone: Regarding the case when there is no rule that has empty URI, implemented finishing the loop when there is no matching Catching Rule.
Regarding more than one "Catch-all" rules I mention to the 1st constraint https://github.com/StarcounterApps/Website/issues/58#issuecomment-286901166. I just propose to add validation of rules on saving.

un-tone:
Similar https://github.com/StarcounterApps/Website/issues/63 issue still impacts on **Website**  with this PR. Need to be investigated.

warpech:
@un-tone could you pls add a detailed information to README.md about how Catch-all rules work after this change?

un-tone:
> Similar #63 issue still impacts on Website with this PR. Need to be investigated.

Fixed in https://github.com/StarcounterApps/Website/pull/65/commits/b712373da24bac96faba9faaba396c0726c74e37

un-tone:
The issue is fixed. `README.md` is updated. Please review.

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -132,6 +134,10 @@ protected void RegisterFilter()
                             wrapper = response.Resource as WrapperPage;
                             requestUri = wrapper?.WebTemplatePage.Data.Html;
                         }
+                        if (this.CurrentResponse == null)
+                        {
+                            Session.Current.Data = response.Resource as Json;
```
warpech: For me this looks smelly. An app can only have one object assigned to a session. In case of WebsiteProvider, session is already assigned [here](https://github.com/StarcounterApps/Website/blob/b712373da24bac96faba9faaba396c0726c74e37/src/WebsiteProvider/Api/ContentHandlers.cs#L63)

@chrhol can you pls review the commit https://github.com/StarcounterApps/Website/pull/65/commits/b712373da24bac96faba9faaba396c0726c74e37? It is supposed to fix PR #63

warpech:
@chrhol ping, can you pls take a look at https://github.com/StarcounterApps/Website/pull/65#discussion_r108710678?

src/WebsiteProvider/Api/ContentHandlers.cs
chrhol: I'm having a bit of a problem understanding exactly what these lines do. As @warpech states, the session is assigned from another place. Is it to ensure that it's always contains something?

warpech: @un-tone can you explain this line a bit? Or should we have a meeting?

warpech:
@un-tone can you pls add tests?

un-tone:
Tests are added.

src/WebsiteProvider/Api/ContentHandlers.cs
un-tone: This is for case when WebsiteProvider doesn't wrap a response and returned it as-is. I assumed that in this case the Session.Data have to be changed. Otherwise, the responded JSON Patch doesn't contain the complete data and nothing is happened on the page.

test/WebsiteProvider_AcceptanceHelperOne/Helpers/DataHelper.cs
```diff
@@ -6,7 +6,7 @@ namespace WebsiteProvider_AcceptanceHelperOne
 {
     public class DataHelper
     {
-        public void GenerateData()
+        public void SetDefaultCatchingRules()
```
warpech: It is hard to follow how does the catching rules config look like after running this helper.

I think that a helper that deletes all data and creates only the ones used in test would be better. WDYT?

src/WebsiteProvider/Api/ContentHandlers.cs
warpech: Ok so this will be long. The test `GoToOwnPage_NoCatchAllRules_LoadedAsIs` tests:

1. Loading `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne/SetupNoCatchAllRuleTest` to prepare the test setup
2. Going to `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne` - a page that is stateless (does not create a session), but is configured to be wrapped in a surface by WebsiteProvider. WebsiteProvider creates the session.
3. Clicking on a "simple page" link that goes to `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne/SimplePage` - a page that is also stateless and configured to be NOT wrapped in a surface by WebsiteProvider

What happens in lines https://github.com/StarcounterApps/Website/blob/8309d12b240a49f2f364d122d13851b423b1dbc9/src/WebsiteProvider/Api/ContentHandlers.cs#L139-L143 is that WebsiteProvider creates a session of response from a different app, and by some mechanism (which one?) makes the response replace the root of the previous response.

So here are some questions:

1. Should we actually wrap stateless pages? I think yes.
2. What should we do when morphing to a stateless page? Should we force it into the session? Or should we actually force a full page reload to that page? (I suppose the latter)
2. @un-tone please note that going to the page `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne/SimplePage` directly actually does not render it. This is because it does not have a session, which causes it to be ignored by middleware (see [this line in middleware](https://github.com/Starcounter/level1/blob/2d6990034e47906453c8108c6191bdf43aa78959/src/Starcounter/Middleware/PartialToStandaloneHtmlProvider.cs#L97))

warpech:
Continued discussion from https://github.com/StarcounterApps/Website/pull/65#discussion_r111356247:

-----

The test `GoToOwnPage_NoCatchAllRules_LoadedAsIs` tests:

1. Loading `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne/SetupNoCatchAllRuleTest` to prepare the test setup
2. Going to `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne` - a page that is stateless (does not create a session), but is configured to be wrapped in a surface by WebsiteProvider. WebsiteProvider creates the session.
3. Clicking on a "simple page" link that goes to `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne/SimplePage` - a page that is also stateless and configured to be NOT wrapped in a surface by WebsiteProvider

What happens in lines https://github.com/StarcounterApps/Website/blob/8309d12b240a49f2f364d122d13851b423b1dbc9/src/WebsiteProvider/Api/ContentHandlers.cs#L139-L143 is that WebsiteProvider creates a session of response from a different app, and by some mechanism (which one?) makes the response replace the root of the previous response.

@un-tone please note that going to the page `http://localhost:8080/WebsiteProvider_AcceptanceHelperOne/SimplePage` directly actually does not render it. This is because it does not have a session, which causes it to be ignored by middleware (see [this line in middleware](https://github.com/Starcounter/level1/blob/2d6990034e47906453c8108c6191bdf43aa78959/src/Starcounter/Middleware/PartialToStandaloneHtmlProvider.cs#L97))

So here are some questions:

1. Should we actually wrap stateless pages? I think yes.
3. What should we do when morphing to a stateless page? Should we force it into the session? Or should we actually force a full page reload to that page? (I suppose the latter)

@chrhol, @miyconst, WDYT?

un-tone:
> The test GoToOwnPage_NoCatchAllRules_LoadedAsIs performs the following (@un-tone, right?)

Right.

test/WebsiteProvider_AcceptanceHelperOne/Api/MainHandlers.cs
```diff
@@ -1,16 +1,44 @@
...
             Handle.GET("/WebsiteProvider_AcceptanceHelperOne/EmptyJson", () => new Json());
-            Handle.GET("/WebsiteProvider_AcceptanceHelperOne/EmptyPage",
-                () => new Page { Html = "/WebsiteProvider_AcceptanceHelperOne/Master.html" });
+            Handle.GET("/WebsiteProvider_AcceptanceHelperOne/SimplePage",
+                () => new Page { Html = "/WebsiteProvider_AcceptanceHelperOne/SimplePage.html" });
+
+            Handle.GET("/WebsiteProvider_AcceptanceHelperOne/SetDefaultCatchingRules", () =>
+            {
+                DataHelper.SetDefaultCatchingRules();
+                return Response.FromStatusCode(200);
```
un-tone: Fixed.

test/WebsiteProvider_AcceptanceHelperOne/Helpers/DataHelper.cs
un-tone: Agree. It makes tests dependent to the Website default data.

un-tone: Done. Also added checkings for a loaded surface.

un-tone:
I will add comments to the test.

warpech:
Ping @miyconst, @chrhol regarding https://github.com/StarcounterApps/Website/pull/65#issuecomment-293850988

miyconst:
Once again we are hit by out session mechanism. In other web techs that I am familiar is the session is created automatically, and you don't have to care about it. Still, there is an option to forcibly disable it, if needed.

Regarding the questions:

>Should we actually wrap stateless pages? I think yes.

I also think yes. If an app does not require session it still should be possible to wrap it with Website app.

>What should we do when morphing to a stateless page? Should we force it into the session? Or should we actually force a full page reload to that page? (I suppose the latter)

I don't see how we could actually check if we redirect to a session-state or session-less page. The only way to check, is to make a request when no session available, and check if session was created after the response.

Anyway, I think we should bring that page into current session, since the page is wrapped within Website app, and Website app requires session.

warpech:
@chrhol what's your opinion on the discussion from https://github.com/StarcounterApps/Website/pull/65#discussion_r109866310 forward?

chrhol:
I agree with @miyconst points for anwers to the questions. 
 
> I don't see how we could actually check if we redirect to a session-state or session-less page.

> Anyway, I think we should bring that page into current session, since the page is wrapped within Website app, and Website app requires session.

Since the website app requires session, everything wrapped in it should be brought in. 

Two additional points. 
- The code mentioned in a comment above (and is quoted underneath) is unsafe to use and there will be problems when more than one request is handled at the same time (I'm not sure that it belongs to this PR though).  Described in #72 
> What happens in lines https://github.com/StarcounterApps/Website/blob/8309d12b240a49f2f364d122d13851b423b1dbc9/src/WebsiteProvider/Api/ContentHandlers.cs#L139-L143 is that WebsiteProvider creates a session of response from a different app, and by some mechanism (which one?) makes the response replace the root of the previous response.
- Hopefully we can come to a decision on a way forward around how `Session` is working which probably will improve the code here for adding and storing viewmodels on the session (https://github.com/Starcounter/level1/issues/4159)

un-tone:
> The code mentioned in a comment above (and is quoted underneath) is unsafe to use

Replaced `CurrentResponse` from the condition with local variable

warpech:
Thanks @un-tone! I believe that `CurrentResponse` and `CurrentTemplate` are code smells and should be refactored to function parameters or view-model properties. Would you have some idea how to improve that?

----

Regarding the main topic of the doubtful lines https://github.com/StarcounterApps/Website/blob/8309d12b240a49f2f364d122d13851b423b1dbc9/src/WebsiteProvider/Api/ContentHandlers.cs#L139-L143, I propose to change them into:

```cs
                        if (!isWrapped)
                        {
                            //Morph to a view that is stateless and not catched by any surface
                            //This is tested by RedirectToOtherAppPageTest
                            //Should be improved by https://github.com/Starcounter/level1/issues/4159
                            Session.Current.Data = response.Resource as Json;
                        }
```

Notice that I have removed the following part:

```cs
                            if (Session.Current == null)
                            {
                                Session.Current = new Session(SessionOptions.PatchVersioning);
                            }
```

My logic behind this is that if we are morphing **out**, the session will always exist. All the tests are still passing. @un-tone, any flaws in my thinking here?

un-tone:
Please review last changes.

src/WebsiteProvider/Helpers/Storage.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Concurrent;
+
+namespace WebsiteProvider
+{
+    public class Storage<T>
+    {
+        private readonly ConcurrentDictionary<string, T> responses = new ConcurrentDictionary<string, T>();
```
warpech: What are the benefits of `ConcurrentDictionary` vs a `Dictionary`? Should we always use `ConcurrentDictionary`? Then let's add it to our guidelines.

PR #68, opened 2017-03-17
Batch scripts refactoring:

KmaTTi89:
- removed make.bat, restart.bat
- fixed build.bat, package.bat, run.bat


build.bat
```diff
@@ -3,6 +3,5 @@
 :: Set up the env to use Msbuild 14.0
 CALL "%VS140COMNTOOLS%\vsvars32.bat"
 
-PUSHD %~dp0
 msbuild
-POPD
+POPD
```
warpech: Why removing PUSHD, POPD? It is needed, otherwise you cannot call `build.bat` from an external directory.

warpech: Please stick to https://starcounter.gitbooks.io/guidelines/content/batch-files.html wherever possible

KmaTTi89: Do we need to go inside tools folder when we don't need to use nuget.exe and this folder doesn't exists?

warpech: No, we don't need to go to the tools folder (we don't need `PUSHD %~dp0\tools`).

But we need `PUSHD %~dp0` because it sets working directory for msbuild.

KmaTTi89: Fixed

warpech: Also, it is wrong to call POPD without earlier calling PUSHD (they both form a stack)

PR #75, opened 2017-04-07
Added Teamcity.ps1 script for CI

KmaTTi89:


warpech:
The TeamCity scripts in all of our apps are very similar: https://www.diffchecker.com/ryRDYfkK

Have you thought of parameterizing the it to make it possible to have exactly the same script in all of our projects?

I would just take a list of paths to `exe` files to run and a list of test `dlls` to provide to the nunit console runner.

PR #80, opened 2017-04-13
Make partial views Shadow DOM v1 ready

tomalec:
Fixes #74
#42

I have also 
 - renamed slot names, and `partial` to `view-model`
 - fixed sequential navigation order of settings icon to match visual one
 - added hover effect for links in top left corner

Issue #79, opened 2017-04-13
Replace AppHub template with HolyGrail template


warpech:
The AppHub template present in the Website app is an example of doing a web site interface with the Website app.

The AppHub template is not flexible and hard to maintain. It should be replaced with more flexible and recognizable HolyGrail template.

See: https://en.wikipedia.org/wiki/Holy_Grail_(web_design)

warpech:
Closing as done

PR #82, opened 2017-04-13
Replace AppHub with HolyGrail layout

tomalec:
I have also fixed standalone navigation layout as it was using floats.

It uses CSS Grid Layout new one and `-ms-grid` one so it works in IE10+, Edge 10+, Firefox 52+, Chrome 57+, Safari10.1+(not tested), Opera 44+, iOS Safari 10.3+(not tested)

This PR requires https://github.com/StarcounterApps/Website/pull/80


warpech:
Implements #79

src/Website/wwwroot/Website/css/holygrail.css
```diff
@@ -0,0 +1,44 @@
...
+    grid-template: "head head head"
+                   "left content right" 1fr
+                   "foot foot foot";
+}
+.website-holygrail__header {
+    -ms-grid-row: 1;
+    -ms-grid-column: 1;
+    -ms-grid-column-span: 3;
+
+    grid-area: head;
```
warpech: Isn't it more maintainable to do:

```css
-ms-grid-row: 1;
-ms-grid-column: 1;
-ms-grid-column-span: 3;
grid-row: 1;
grid-column: 1;
grid-column-span: 3;
```

Then it becomes apparent that we are expecting the same in Edge and other browsers

warpech:
@tomalec can you pls merge the current develop to this branch?

src/Website/wwwroot/Website/css/holygrail.css
tomalec: I would like not to maintain per-column and row layout at all. I thought that it would be easier to just remove l.14-16 once IE catch up, without a need to migrate from x,y coords to grid areas.

warpech: Looking at the release cycle, it might take 9-12 months until EdgeHTML 16 that allegedly should support CSS Grid Layout final spec. So we *do have* to maintain the per-column and row layout. Question is do we also have to support template-based layout.

tomalec: Both are "template-based layout" as `-ms-grid-column` ~= `grid-template-columns`.  `grid-template` is a shorthand. So we are speaking about areas-based layout.

(Also there is no `grid-*-span`, there is `grid-*: 1/ span 3`)

Making them both use  x,y coordinates will suggest the both layouts works the same, which is not entirely true, as old `-ms-grid` works slightly different in regards of track sizing and intrinsic sizes. (See that track sizes in grid and `-ms-grid` are different). 

Therefore, per-area it would become apparent that we are expecting the same thing. But per-grid-template even though we expect same thing, we need to define it differently and more explicitly for `-ms` what would become slightly  less apparent.

Personally I find maintenance of named areas many times easier. Then with IE support added on a side we could maintain it like:
1. Change `grid-template` areas.
2. Search in the code for change areas, for example `header, footer`.
3. Adjust x,y for `-ms-grid-row|column`.

Otherwise, you would have to:

1. Change `grid-template-rows`, `grid-template-rows`.
2. Calculate in your own head which areas are to be changed and to which values.
2. Search in the code for them **manually**, as searching for `2, 3` does not make much sense.
3. Adjust x,y for `-ms-grid-row|column`.

warpech: Wow, I'm convinced :)

warpech:
Maybe it would be good to add some explicit text to every section in the layout? And more distinct coloring, like here: http://ireade.github.io/holy-grail-css-grid/

Responsiveness would actually be good, too!

tomalec:
Do you think so? Should that text persist once something is distributed, or me just a fallback for no content?

Coloring and hardcoding text, in my opinion it would make the layout unusable on production.

tomalec:
By responsive holy grail, you mean changing it to stack? It would no longer by holy grail, but sure I can do this.

tomalec:
Speaking of responsiveness, I think Default and Launcher should use grid and responsively change it to stack if needed, as they do not impose by name any particular layout.

warpech:
My reasoning is that HolyGrail should be more of a learning tool than production ready, so it can have:

- section labels (fallback content, or persistent - I leave it up to you, I would prefer persistent)
- section colors
- responsive media queries
- text saying that you can modify it by editing its HTML composition

WDYT?

tomalec:
If it's just a learning tool I can agree on section labels and colors, as long as we have some disclaimer visible.

Speaking of labels, I'd prefer to use fallback content, as this would be yet another thing to showcase - using of fallbacks. Unfortunately it does not work, as Websites always inserts (sometimes empty) another WebSection  object.

src/Website/wwwroot/Website/templates/LauncherTemplate.html
```diff
@@ -293,7 +255,7 @@
 <dom-module id="website-launchertemplate-menu">
     <template>
         <style></style>
-        <content id="content"></content>
+        <content id="content"></slot>
```
un-tone: Please fix `<content `

warpech: Good catch! Actually it was released unnoticed in previous PR: https://github.com/StarcounterApps/Website/commit/e5958cd5600775c84bbda855a053949c158369a6

warpech:
> If it's just a learning tool I can agree on section labels and colors, as long as we have some disclaimer visible.

The colors can be configurable using CSS Custom Properties. Disclaimer is totally fine!

tomalec:
@warpech Should we try to fix fallback content in clean way, or for example try to replace it with CSS `content` with Custom Properties?

src/Website/wwwroot/Website/templates/LauncherTemplate.html
tomalec: fixed

warpech:
> Should we try to fix fallback content in clean way, or for example try to replace it with CSS `content` with Custom Properties?

Any option sounds nice! Try something that appeals to you the most

Issue #64, opened 2017-03-14
Transaction state is not always being updated on the client side


un-tone:
### Steps to reproduce
1. Start **Website** (branch `develop`), **WebsiteProvider** (branch `develop`), **SignIn** (branch `develop`)
2. Go to `http://localhost:8080/website/cms/catchingrules`
3. Change something

### Expected result
"Save" button get active.

### Actual result
Nothing happened with the "Save" and "Cancel" buttons on the form. The client side doesn't get changes of the current transaction state from the backend.
_**Note:** sometimes it works correctly, in this case try to refresh the current page._

##### Involved apps' version
 - StarcounterApps/Website@04fd00046e025d618fdbf89c211b28bc54f4d0dd
 
##### Browsers affected
- [x] Chrome
- [x] Firefox

### Screenshot
![image](https://cloud.githubusercontent.com/assets/1625378/23896834/d515f7e8-08bb-11e7-8602-6e70a8245e13.png)
_For example I reset all "Final" flags._

warpech:
I confirm this also happens to me.

un-tone:
Created PR https://github.com/StarcounterApps/Website/pull/83. Please review.

un-tone:
Closing as done

PR #83, opened 2017-04-13
Removed CommitUpdate hook for SystemUserSession

un-tone:
For https://github.com/StarcounterApps/Website/issues/64.

This change eliminates second redundant request for current page (which
is in scope of the SystemUserSession processing transaction) so on pages
which are works with transactions (e.g. CmsCatchingRulesPage) will be
initialized a correct not-null Transaction represenation


warpech:
Interesting! Looks good to me.

Issue #59, opened 2017-03-09
CompositionEditor only works correctly when started before Website


warpech:
### Steps to reproduce

1. Start Website `run.bat` (`develop` branch)
2. Start CompositionEditor `run.bat` (`develop` branch)
3. Go to `http://localhost:8080/website/cms`
4. Press CTRL+E

### Expected result

The Composion Editor panel should show up.

### Actual result

Nothing happens, the browser address bar gets focused.

This happens because Website middleware (`Application.Current.Use`) cancels CompositionEditor middleware. 

The editor works as expected if you run the apps in the following order:

1. CompositionEditor `run.bat`
2. Website `run.bat`
3. Press CTRL+E

un-tone:
Created PR https://github.com/StarcounterApps/Website/pull/85

un-tone:
Closing as done

PR #85, opened 2017-04-17
Skipping changing requests in the middleware

un-tone:
For https://github.com/StarcounterApps/Website/issues/59

### DIFF
--- a/src/WebsiteProvider/Api/ContentHandlers.cs
+++ b/src/WebsiteProvider/Api/ContentHandlers.cs
@@ -141,8 +141,8 @@ namespace WebsiteProvider
                             wrapper = response.Resource as WrapperPage;
                             requestUri = wrapper.WebTemplatePage.Data.Html;
                         }
+                        return response;
                     }
-                    return response;
                 }
                 return null;
             });
--- a/src/WebsiteProvider/Api/ContentHandlers.cs
+++ b/src/WebsiteProvider/Api/ContentHandlers.cs
@@ -142,8 +142,9 @@ namespace WebsiteProvider
                             requestUri = wrapper.WebTemplatePage.Data.Html;
                         }
                     }
+                    return response;
                 }
-                return response;
+                return null;
             });
         }
 
--- a/test/WebsiteProvider.Tests/Test/PageLoadTests.cs
+++ b/test/WebsiteProvider.Tests/Test/PageLoadTests.cs
@@ -1,6 +1,4 @@
-﻿using System.Linq;
-using NUnit.Framework;
-using OpenQA.Selenium;
+﻿using NUnit.Framework;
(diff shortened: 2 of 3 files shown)


src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -142,8 +142,9 @@ protected void RegisterFilter()
                             requestUri = wrapper.WebTemplatePage.Data.Html;
                         }
                     }
+                    return response;
```
warpech: Are you sure `return response;` should not be before line 144?

Can you pls add tests?

un-tone: Fixed and added test.

test/WebsiteProvider_AcceptanceHelperTwo/Program.cs
```diff
@@ -9,6 +9,24 @@ static void Main()
...
+  var inserter = document.createElement(""input"");
+  inserter.setAttribute(""type"", ""hidden"")
+  inserter.setAttribute(""test"", ""acceptance-helper-two-middleware"")
+  document.body.insertBefore(inserter, document.body.firstChild);
+})();
+</script>";
+                    return new Response { Body = $"{response.Body}\n{inserter}" };
+                }
+
+                return null;
```
warpech: Have you tried communicating with something less hacky, e.g. a cookie? SignIn has middleware that adds a cookie.

A cookie has to be cleared after the test, though.

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
```diff
@@ -38,5 +40,12 @@ public void RedirectToOtherAppPageTest()
             _acceptanceHelperTwoMasterPage = _acceptanceHelperOneMasterPage.GoToAcceptanceHelperTwoPage();
             WaitForText(_acceptanceHelperTwoMasterPage.H1Element, "Acceptance Helper 2", 10);
         }
+
+        [Test]
+        public void LoadPage_MiddlewareHandling_ContentInjected()
```
warpech: I would prefer to add a comment above the test what exactly are we testing here. This applies to other tests, too.

test/WebsiteProvider_AcceptanceHelperTwo/Program.cs
un-tone: Implemented with cookie.

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
un-tone: Added description for this new test. Since the other tests was strongly modified in the https://github.com/StarcounterApps/Website/pull/65 I will add comments there.

Issue #61, opened 2017-03-10
Surface composition is not editable with CompositionEditor


warpech:
### Steps to reproduce

1. Start CompositionEditor `run.bat` (`develop` branch)
2. Start CompositionProvider `run.bat` (`develop` branch)
3. Start Website & WebsiteProvider `run.bat` (`develop` branch)
4. Go to `http://localhost:8080/website/cms`
5. Press CTRL+E
6. Double click on the input in the composition editor and select `[slot="WebsiteProvider/Surface/DefaultTemplate"]`
7. Press "Save" button

### Expected result

The composition should be saved to the db.

### Actual result

A JavaScript error occurs:

```txt
Uncaught TypeError: Cannot set property 'Composition$' of undefined
    at HTMLElement.<anonymous> (starcounter-include.html:287)
```

This happens because `WebTemplatePage` in WebsiteProvider is not requested through `Self.GET`.

### Screenshot

![2017-03-10 13_40_10-website](https://cloud.githubusercontent.com/assets/566463/23795839/ede660d6-0597-11e7-9ee2-f7b48f1051b6.png)

un-tone:
Created PR https://github.com/StarcounterApps/Website/pull/86

un-tone:
Closing as done.

PR #86, opened 2017-04-17
Moved creating of WebTemplatePage to Handle.GET

un-tone:
For https://github.com/StarcounterApps/Website/issues/61


un-tone:
I suppose it is enough to test on regression. I've tested with tests from https://github.com/StarcounterApps/Website/pull/65. WDYT

warpech:
Sounds fine, thanks!

PR #92, opened 2017-04-25
Extended URI wildcard

un-tone:
For https://github.com/StarcounterApps/SignIn/issues/89


src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -11,13 +11,17 @@ public class ContentHandlers
...
 
-            Url = reg.Replace(Url, "/{?}");
+            if (reg.IsMatch(url))
+            {
+                return reg.Replace(url, "/{?}");
+            }
 
-            return Url;
+            reg = new Regex(@"[?](\w|\d|%)*$", RegexOptions.IgnoreCase);
+            return reg.Replace(url, "?{?}");
```
warpech: I think that it is too much guessing. Is it possible to do the other way round? Instead of trying to guess the Handler from URL, could we check the registered handlers and check if the URL parameter matches any it?

That's a question to: @alemoi

alemoi: What's the end goal here? To find the matching handler for a given URI (where parameters are given from the browser)?

warpech: I think so. @un-tone, right?

alemoi: As I understood he converts the actual URI into a handler URI. Correct @un-tone? And then he needs to check if its registered via `Handle.IsHandlerRegistered`

warpech: I want to do the opposite check. Get a list of registered handlers (in all apps) and check one by one if `url` matches any of them.

un-tone: In this method I convert actual URI into a [Catch URI](https://github.com/StarcounterApps/Website/blob/develop/README.md#catching-rules) which is used to find a corresponding Catching Rule. I don't see a reason to call the `Handle.IsHandlerRegistered`.

I suppose it would be helpful to use something like this:
```
string uriTemplate = Handle.GetCorrespondingUriTemplate(actualUri);
WebUrl catchingRule = Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url = ?", uriTemplate).First;
```

warpech: By `Handle.GetCorrespondingUriTemplate` we mean a registered handler pattern like `/people/person/{?}` that matches the given uri `/people/person/123`.

warpech:
@un-tone can you pls do the following:

1. Add tests for this change
2. Add mention to README.md about supported Catch URI patterns

Then we can merge the PR and make a follow up issue about improving the `GetWildCardUrl` method.

src/WebsiteProvider/Api/ContentHandlers.cs
alemoi: > I want to do the opposite check. Get a list of registered handlers (in all apps) and check one by one if url matches any of them.

Would you like me to add something `Handle.GetRegisteredHandlers()` as a list of strings or dictionary with a format "method_space_uri"

un-tone:
Added tests and updated README. @warpech please review

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace WebsiteProvider.Tests.Test
 {
-    [Parallelizable(ParallelScope.Fixtures)]
+    [Parallelizable(ParallelScope.None)]
```
warpech: What's the problem with that?

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
```diff
@@ -19,6 +19,19 @@ public PageLoadTest(Config.Browser browser) : base(browser)
         {
         }
 
+        // TODO : Remove these 2 methods when merging with the PR Website#65
+        [OneTimeSetUp]
+        public void SetUpFixture()
+        {
+            Driver.Navigate().GoToUrl("http://localhost:8080/website/resetdata");
```
warpech: I that some tests call `http://localhost:8080/website/resetdata` while others call `Config.AcceptanceHelperOneUrl + "/ResetData"`.

Maybe it's better to only call `Config.AcceptanceHelperOneUrl + "/ResetData"` and make that one use `Website.DataHelper.ClearData` internally?

src/WebsiteProvider/Api/ContentHandlers.cs
warpech: As discussed with @alemoi on HO, the above is desirable, but even better and also possible is to add a function to just match a given URI and return a handler URI.

Created follow up issues for that: https://github.com/Starcounter/Home/issues/99, https://github.com/StarcounterApps/Website/issues/94

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
un-tone: When I made changes in this class I relied on the changes from the https://github.com/StarcounterApps/Website/pull/65 where the PageLoadTests is completely implemented. I just don't want to do this work twice and I'm going to clean the changes from this PR when merging with that changes. Please tell me if I didn't persuaded you to stay this changes as-is.

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
un-tone: These tests rely on the Website data which is different for different fixtures and tests. One test case can be incompatible with another one so some tests can't be passed.

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
warpech: I think that is fine. Can you please make an issue about that so we don't forget?

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
warpech: Makes sense. Can you add a comment about that in the test?

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
un-tone: Made https://github.com/StarcounterApps/Website/issues/96

test/WebsiteProvider.Tests/Test/PageLoadTests.cs
un-tone: Added.

Issue #105, opened 2017-06-02
Cleanup wrapping


tomalec:
It seems that Website does some magic build on top of the buggy fact we have used `imported-template` on root level https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteProvider/ViewModels/WrapperPage.json.cs#L6-L8
> 
    ///  WrapperPage needs to wrap WebTemplatePage, because Starcounter applies middleware
    ///  from the original request. The default middleware contains &lt;template is="imported-template"&gt;&lt;/template&gt;
    ///  instead of &lt;starcounter-include&gt;, which is why we have to wrap it.

Since https://github.com/Starcounter/level1/pull/4223, that is no longer true

##### Involved apps' version

 - StarcounterApps/Website@162d13c9263ab8e47e0471c1759951445fe0f2c6

warpech:
Implemented in https://github.com/StarcounterApps/Website/pull/107

warpech:
Merged

PR #107, opened 2017-06-06
Refactor WebsiteProvider's WrapperPage

warpech:
As proposed by @tomalec in #105, starting from Starcounter 2.3.1.6523 `starcounter-include` is inserted automatically by middleware so there should be no need for WrapperPage in WebsiteProvider.

This PR removes WrapperPage.

@un-tone pls review. There might be some dragons hiding in the closet. All automated tests and my manual tests seem to be ok, though.

As a drawback, CompositionEditor now cannot save changes in the topmost surface. This is dependent on https://github.com/Starcounter/level1/issues/4017


warpech:
The build is failing on TC. It seems that is some cleanup problem on the TC: https://teamcity.starcounter.org/viewLog.html?buildId=25626&tab=buildResultsDiv&buildTypeId=Apps_Website_Develop

I have asked @joozek78 for help.

warpech:
Ping @joozek78, @alemoi. Can you pls help? The builds in PRs of the Website app are failing for no apparent reason.

alemoi:
I have changed now the Clean policy to Always. Maybe that helps.

joozek78:
@alemoi Where did you change it? For this build configuration? Project? Globally?

warpech:
Thanks! Trying to re-run

joozek78:
BTW you could just re-run with this one build forced to clean

alemoi:
In VCS settings, applied to all projects: 

image

Lets see if it helps.

warpech:
> BTW you could just re-run with this one build forced to clean

I was trying this two days ago, with no help...

@alemoi your change didn't help either, still failing: https://teamcity.starcounter.org/viewType.html?buildTypeId=Apps_Website_Develop&branch_Apps_Website=simpler-provider&tab=buildTypeStatusDiv

joozek78:
@alemoi can you revert the change? since it didn't help, we don't want it to last - it can slow down further builds

joozek78:
This isn't about TC, it fails on my machine too

---
Pozdrawiam,
Marcin Goliński

warpech:
Sorry, I have just noticed too. I guess the problem is on **me**.

joozek78:
Take a look [here](https://github.com/starcounterprefabs/website/blob/9a9134423741371455cacc0008ab646b047b4873/src/WebsiteProvider/ViewModels/SectionPage.json#L5) - it seems this class doesn't exist (maybe renamed?)

alemoi:
Yes, seems the problem is with something else.

warpech:
Seems that a dog ate half of my refactoring. Sorry guys for trouble. I must fix it in my code.

warpech:
Fixed with a force push. @un-tone could you pls review?

src/WebsiteProvider/ViewModels/SurfacePage.json.cs.orig
```diff
@@ -1,21 +1,22 @@
+using Simplified.Ring6;
 using Starcounter;
 
 namespace WebsiteProvider
 {
+<<<<<<< HEAD
```
joozek78: that doesn't seem good

warpech: I should stick to SourceTree, not the command line :|

Issue #101, opened 2017-05-30
Pinning rule is not unmapped correctly


warpech:
### Steps to reproduce

1. Start Website (branch `develop`), WebsiteProvider (branch `develop`), SignIn (branch `develop`)
2. Call `http://localhost:8080/website/resetdata` to load default mappings
3. Go to `http://localhost:8080/signin/generateadminuser` to generate the admin user
4. Go to `http://localhost:8080/website/cms`
5. Sign in as admin user
6. Go to `http://localhost:8080/website/cms/pinningrules`
7. Change the row that says `/signin/user` to `/signin/user2`
8. Press "Save"

### Expected result

The "Sign In" icon should disappear from the top right corner.

### Actual result

The "Sign In" icon does not disappear from the top right corner. This probably means that the pinning rule is not unmapped correctly in WebsiteProvider.

##### Involved apps' version

 - StarcounterApps/Website@d3e2d04bce285cc1bb66b6e820d446a81e301254
 - StarcounterApps/SignIn@8c10b2f1c716bade691a33beee5128cff2a33618
 
##### Browsers affected

- [x] Chrome

### Screenshot

![2017-05-30 15_52_13-website](https://cloud.githubusercontent.com/assets/566463/26586610/feb6734a-454f-11e7-9662-38c6acc468ec.png)

warpech:
@un-tone is this perhaps fixed by PR #111?

un-tone:
Unmapping a rule is not implemented for rules changing, it works only for rules creating/deleting. Sorry that I haven't notified about it explicitly.

PatrykSzwer:
Done in #121. Closing.

PR #121, opened 2017-06-26
101 unmap pinning rule

PatrykSzwer:
#101

### DIFF
--- a/src/WebsiteProvider/Api/MappingHandlers.cs
+++ b/src/WebsiteProvider/Api/MappingHandlers.cs
@@ -57,16 +57,13 @@ namespace WebsiteProvider
 
         public void UpdatePinningRule(WebMap webMap)
         {
-            var newUri = webMap.ForeignUrl;
-            webMap.ForeignUrl = GetOldUri(webMap);
-            this.UnmapPinningRule(webMap);
-
-            webMap.ForeignUrl = newUri;
+            this.UnmapPinningRule(webMap, GetOldUri(webMap));
             this.MapPinningRule(webMap);
         }
 
-        public void UnmapPinningRule(WebMap webMap)
+        public void UnmapPinningRule(WebMap webMap, string webMapForeignUrl = null)
         {
+            webMapForeignUrl = webMapForeignUrl ?? webMap.ForeignUrl;
             if (webMap.Section?.Template == null)
             {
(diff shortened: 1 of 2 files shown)


src/WebsiteProvider/Api/MappingHandlers.cs
```diff
@@ -54,13 +55,24 @@ public void MapPinningRule(WebMap webMap, string registeredSectionUri = null)
             Blender.MapUri(webMap.ForeignUrl, token);
         }
 
+        public void UpdatePinningRule(WebMap webMap)
+        {
+            var newUri = webMap.ForeignUrl;
+            webMap.ForeignUrl = GetOldUri(webMap);
```
un-tone: From my point of view it is not a good practice to make changes in DB objects just to pass some value to a method. In current case I recommend to add additional parameter, e.g. `string webMapForeignUrl = null` and use it in called methods like this `webMapForeignUrl ?? webMap.ForeignUrl`.

PatrykSzwer: That's actually the idea I was looking for! Thanks!

src/WebsiteProvider/Api/MappingHandlers.cs
```diff
@@ -54,17 +55,25 @@ public void MapPinningRule(WebMap webMap, string registeredSectionUri = null)
...
-        public void UnmapPinningRule(WebMap webMap)
+        public void UpdatePinningRule(WebMap webMap)
         {
+            this.UnmapPinningRule(webMap, GetOldUri(webMap));
+            this.MapPinningRule(webMap);
+        }
+
+        public void UnmapPinningRule(WebMap webMap, string webMapForeignUrl = null)
+        {
+            webMapForeignUrl = webMapForeignUrl ?? webMap.ForeignUrl;
```
un-tone: A program should not do anything redundant. This line should be moved below after the condition block.

PatrykSzwer: Done

un-tone:
Please don't forget to create a new issue for tests covering this functionality.

PatrykSzwer:
I've created an issue for tests here: https://github.com/StarcounterApps/Website/issues/126

Issue #109, opened 2017-06-07
Implement new UI


warpech:
Website becomes a key app but it has a very obsolete UI. I would like to discuss the UI on a meeting at 2017-06-14.

My inspiration for the new UI is Cloudflare's admin panel: 

- https://youtu.be/HsqCB3K9IN8?t=24s
- https://youtu.be/51PXQMDfcFs?t=52s

Of course, in our case the front page does not select domain names but Website surfaces. Everything else happens in context of the selected surface.

warpech:
@PatrykSzwer is working on designing the new UI for the WebsiteEditor. It can be previewed here: https://preview.uxpin.com/5fa3f26d8d1b7794234e9d1cf0f421ffd1f2ec38. To implement this, we will need first to split the view-models to the desired ones (but keep them ugly), then do the HTML, creation of surface groups. It needs to be done by next week, so we can start working on the enhancements.

PatrykSzwer:
Done, closing.

PR #129, opened 2017-07-07
109 Creating new UI - rearranged views.

PatrykSzwer:
As reference to #109 this is first step of implementing new UI. Views were rearranged to concept of single surface scope. All features from old UI should be reachable from new one. This step opens possibility for working simultaneously on different views without conflicting with each other.


src/WebsiteEditor/Helpers/IKnowSurfacePage.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace WebsiteEditor.Helpers
```
PatrykSzwer: Since day I first saw this It kinda bothers me. I'll make It in an old way, since I agree its subject for separated issue.

src/WebsiteEditor/Helpers/IKnowSurfacePage.cs
```diff
@@ -0,0 +1,9 @@
...
+{
+    public interface IKnowSurfacePage
+    {
+        string SurfaceKey { get; set; }
+
+        void RefreshData();
```
PatrykSzwer: I'm not sure about adding parameter to RefreshData. This method is used not only at the beginning by mainHandlers but also later, inside class itself, by e.g Handle on CancelChanges in PinningRulesPage. There I would need to have surfaceKey variable to pass it to refreshData so it would end up with similar solution, with creating global variable and setting it before. I agree with checking for null.

un-tone: How about just throwing HTTP 404 status code?

PatrykSzwer: I've added entire errorPage for now and future usage. Please take a look.

src/WebsiteEditor/ViewModels/ErrorPage.json.cs
```diff
@@ -0,0 +1,9 @@
+using Starcounter;
+
+namespace WebsiteEditor.ViewModels
```
un-tone: Should be just `WebsiteEditor` namespace

src/WebsiteEditor/Api/MainHandlers.cs
```diff
@@ -114,7 +138,14 @@ public void Register()
                     master.ShowNavigation = true;
                     master.Surface.Data = Db.SQL<WebTemplate>("SELECT t FROM Simplified.Ring6.WebTemplate t WHERE t.Key = ?", key).First();
 
-                    master.RefreshCurrentPage("/WebsiteEditor/partials/pinningrules");
+                    try
```
un-tone: Isn't it better to extract this peace of code and the similar ones above into some `private RefreshCurrentPage(string partialUri)` method? It's just a proposal from me.

src/WebsiteEditor/Api/MainHandlers.cs
```diff
@@ -159,17 +190,13 @@ protected void RegisterPartials()
             {
                 CatchingRulesPage page = new CatchingRulesPage();
 
-                page.RefreshData();
-
                 return page;
             });
 
             Handle.GET("/WebsiteEditor/partials/pinningrules", () =>
```
un-tone: Let's simplify it to `Handle.GET("/WebsiteEditor/partials/pinningrules", () => new PinningRulesPage())`. The same with the handlers above and below please.

PatrykSzwer: Was already doing it, right after I saw what left without that refresh 😃

PatrykSzwer:
Besides changes I also removed unused anymore surfacesPage files. @un-tone please take a look now.

Issue #118, opened 2017-06-22
Roadmap


PatrykSzwer:
This represents the Roadmap for Website app changes, from bug fixes to visual changes.

### Roadmap

1. [x] 1. Refactoring and initial changes
    1. [x] 1. ~~Rename the app "Website" to "WebsiteEditor #34~~
    2. [x] 2. ~~Rename "*Template" to "*Surface" #114~~
<p>

2. [x] 2. Investigation and fixing of discovered bugs
    1. [x] 1. ~~Save button does not get enabled for a newly signed in user. #102~~
    2. [x] 2. ~~Pinning rule is not unmapped correctly #101~~
    3. [x] 3. ~~Signup icon is not visible with Latest SC RC (2.3.1.6410) #95 (Need confirmation)~~
    4. [x] 4. ~~Cannot change parent of objects in Typed JSON trees #88 (Need further explanation)~~
    5. [x] 5. ~~Website Provider is detected by Avast antivirus as threat #76~~
    6. [x] 6. ~~NullReferenceException when Pinning Rule refers to a static html #56~~
    7. [x] 7. ~~Moving to a page with not final Catching rule reloads whole Surface #55~~
    8. [x] 8. ~~Remapped URLs should appear only once #14 (Out of scope)~~
<p>

3. [x] 3. ~~Release of new, bug-free version and deployment at inside.starcounter.com~~
4. [x] 4. ~~2.3.0 Starcounter release support - Fix on IsHandlerRegistered usage #108 (Done, waiting for merge)~~
<p>

5. [x] 5. Visual changes and enhancements
    1. [x] 1. ~~Discuss and design new UI for Website #109~~
    2. [x] 2. ~~Implement new UI navigation and separate views based on one surface scope concept #109~~
    3. [x] ~~3. Merge Pinning rules view with blending points as accordion/collapse rows. #109~~
    4. [x] ~~4.  Support groups of surfaces #122~~
<p>

6. [x] ~~6. Release of new, visually changed version and deployment at inside.starcounter.com~~
7. [x] ~~7. Allow more catching rules than just URI #48~~
8. [x] ~~8. Release of new version with more catching rules feature and deployment at inside.starcounter.com.~~
9. [x] ~~9. Add possibility to move surfaces between groups.~~
10. [x] ~~10. Deleted pinning rules get unmapped before commiting transaction #140~~
11. [x] ~~11. Error when saving a pinning rule with same properties #139~~
[...]

PR #135, opened 2017-07-13
109 new ui 2

PatrykSzwer:
#109, point 5.3 of roadmap #118.
It's not mergable yet, I have still some small cleaning to do. Functionality itself can be reviewed and tested.
I merged pinning rules into blending points view, also moved default blending point option to general subpage.


PatrykSzwer:
Ok, @un-tone, if You can please review. I think it's mergable now.

un-tone:
I started to review. First thing - 1 test failed:
image
@PatrykSzwer please take a look

un-tone:
It's strange but after running tests in the `develop` branch, in the current it works well too. @PatrykSzwer, nevermind.

src/WebsiteProvider/Api/MappingHandlers.cs
```diff
@@ -25,6 +25,11 @@ public void Register()
 
         public void MapPinningRule(WebMap webMap, string registeredSectionUri = null)
         {
+            if (webMap.Section == null)
```
un-tone: I think this case should be unacceptable and the method should throw an exception when this condition returns true. User should not be able to save such pinning rules. WDYT?

src/WebsiteProvider/Api/MappingHandlers.cs
```diff
@@ -71,7 +76,10 @@ public void UnmapPinningRule(WebMap webMap, string webMapForeignUrl = null)
             string token = webMap.GetMappingToken();
             string mapUri = webMap.GetMappingUrl();
 
-            Blender.UnmapUri(webMapForeignUrl, token);
+            if (webMapForeignUrl != null)
```
un-tone: I think this case should be unacceptable too. Please look at the comment above

PatrykSzwer:
About that: 
`There are methods which delete all WebsiteEditor data: WebsiteEditor.DataHelper.ClearData() and WebsiteProvider_AcceptanceHelperOne.DataHelper.ResetData(). You should add there clearing of new WebGroup data`
It's not in scope of those changes. I'll add that in next PR/branch with surface groups.

src/WebsiteProvider/Api/MappingHandlers.cs
PatrykSzwer: If changed the one above but this one I left unchanged. It's because right now I'm using it when someone is trying to save empty pinning rule (second bug). Maybe You have some good alternative over that? Changes with fix are pushed.

warpech:
The following style makes it very hard to locate inputs, because they appear borderless:

```css
.website-table__control {
    box-shadow: none !important;
    border-bottom: none;
}
```

is there an apparent reason for it? If not, I vote for removing it

src/WebsiteProvider/Api/MappingHandlers.cs
un-tone: I suppose you could remove this condition after fixing the second bug, right?

un-tone:
The app cannot be built in last SC version. Please update from `develop` after merging PR https://github.com/StarcounterApps/Website/pull/137.

PatrykSzwer:
Yes of course, I'll do it right after You merge #137.

PatrykSzwer:
Ok, It's merged to current develop.

Issue #48, opened 2017-03-01
Allow more catching rules than just URI


warpech:
For the beginning, it should be possible to use not just URI for a rule but also a regexp matched against one or more HTTP headers (`User-Agent`, `Cookie`, etc).

CloudFlare has a nice UI for that. Check out the video: https://www.cloudflare.com/features-page-rules/

@un-tone WDYT about that?

un-tone:
@warpech looks nice

warpech:
We need the following kinds of the catching rules: 

- URI path matches (this one already exists). Intended for the `path` fragment of the URI that starts after the domain name
- HTTP header matches. Intended for matching the value of such HTTP headers as: `host`, `user-agent`, etc

The rules specified for a single surface should be treaded as "AND" rules (not "OR").

warpech:
@bouguila could you please work on this? Recommended reviewer: @un-tone

un-tone:
Created PRs https://github.com/StarcounterApps/Website/pull/136 and https://github.com/StarcounterApps/Simplified/pull/62

un-tone:
Closing as done

Issue #115, opened 2017-06-21
Separate "catch all" from Catching Rules


warpech:
"Catch all" right now is an empty URI in Catching Rules.

This makes it impossible to create a specific Catching Rule for URLs like: `http://starcounter.com`, because this indeed is literally an empty URI in our case. 

"Catch all" should be a boolean property of Catching Rule.

warpech:
Maybe we just need to keep catch-all rules on the Catching Rules page, because I want to specify the following rule:

- catch any URL that has **subdomain.mydomain.com** in the `Host` header

warpech:
This is dropped in favor of https://github.com/StarcounterApps/Website/issues/150

PR #136, opened 2017-07-15
Enhance catching rules

un-tone:
For https://github.com/StarcounterApps/Website/issues/48.
Related PR: https://github.com/StarcounterApps/Simplified/pull/62.


warpech:
I am missing one thing. In the new UI where do I set the **catch-all* rule for a surface?

The new feature of enhanced catching rules needs to be applicable for **catch-all** rules too, so that I can say that a particular surface catches all requests that have **mobile** in the `User-Agent` header. Or requests that have **subdomain.mydomain.com** in the `Host` header.

un-tone:
@warpech you can apply headers for catch-all rules in the same way as for other rules: create a rule with an empty URI, add a header for the rule (header key is case-insensitive) and save it.
In the last commit I made some fixes for request handling, please try to check this feature again.

un-tone:
I'm a little confused. Maybe you expected that for catch-all rules there will be some separated form area and elements in UI?

warpech:
> I'm a little confused. Maybe you expected that for catch-all rules there will be some separated form area and elements in UI?

Yes, sorry. I wrongly assumed https://github.com/StarcounterApps/Website/issues/115 is included in  https://github.com/StarcounterApps/Website/pull/135, but it is not.

un-tone:
@PatrykSzwer the conflict is resolved and the PR is re-targeted to `develop`.

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -104,19 +110,34 @@ protected void RegisterFilter()
                         var wrapper = response.Resource as SurfacePage;
                         var requestUri = request.Uri;
                         var isWrapped = false;
+                        var requestKey = RequestStorage.Put(request);
 
-                        while ((wrapper == null || wrapper.IsFinal == false) && this.HasCatchingRule(requestUri))
+                        try
```
PatrykSzwer: You're using try here but without any exception catching. What about exceptions that can came up here? Shouldn't those be properly handled, not throwing without catching like that and forgot?

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/CatchingRulesPage.html
```diff
@@ -1,57 +1,85 @@
...
                         Cancel
                     </button>
-                    <button type="button" class="btn btn-xs btn-default" value="{{model.Create$::click}}"
+                    <button type="button" class="btn btn-xs btn-default" value="{{model.CreateTrigger$::click}}"
                             onmousedown="++this.value;" title="Create catching rule">
                         Create
                     </button>
                 </span>
-                Catching rules
+                Catching rules for {{model.SurfaceName}}
```
PatrykSzwer: I can't see need for telling user again in which surface scope is he right now. I think information on the navigation bar is enough and this is redundant. I would leave just "Catching rules" title.

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -221,24 +242,45 @@ protected SurfacePage GetSurfacePage(WebTemplate template)
             return page;
```
PatrykSzwer: The method `WrapExternalRequest` is not used anymore any should be removed (lines 205-209, couldn't made comment specifically on them).

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -221,24 +242,45 @@ protected SurfacePage GetSurfacePage(WebTemplate template)
...
+        private WebUrl FindUrlByHeaders(QueryResultRows<WebUrl> webUrls, Dictionary<string, string> requestHeaders)
+        {
+            if (!webUrls.Any())
+            {
+                return null;
+            }
+            if (webUrls.All(x => !x.Headers.Any()))
+            {
+                return webUrls.First();
+            }
```
PatrykSzwer: I have a question about this part. I don't really get it, in method for finding Url, when there is no any weburl with headers u're returning first one of them? Could You explain to me why this is like that and how should It work?

src/WebsiteProvider/Api/ContentHandlers.cs
un-tone: Here I need only in `finally` block. Currently I suppose it is an optimal variant to just throw an unhandled exceptions here. If you have any proposals how we should handle these exceptions we could discuss it in separate issue, WDYT?

src/WebsiteProvider/Api/ContentHandlers.cs
un-tone: Removed, thanks

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/CatchingRulesPage.html
un-tone: Removed it.

src/WebsiteProvider/Api/ContentHandlers.cs
un-tone: Yes, it is how it worked before this changes:
`Db.SQL<WebUrl>("SELECT wu FROM Simplified.Ring6.WebUrl wu WHERE wu.Url = ?", url).First`. We just get first appropriate catching rule for current request.
URI of catching rule has a higher priority than headers so if current WebUrl query doesn't have rules with headers then we return the one found only by URI.

un-tone:
I can't see the reason of the TC build failure. @warpech could you take a look at this, please.
It works well on my local machine.

warpech:
TeamCity is getting this error when resolving Nuget:

```
[16:09:34][restore] The 'Newtonsoft.Json 10.0.3' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.60717.93'.
[16:09:34][restore] 'NUnit' already has a dependency defined for 'NETStandard.Library'.
[16:09:34][restore] Process exited with code 1
```

I think this is the same problem as https://github.com/Starcounter/level1/issues/3930#issuecomment-314440227

@mmnosek do you know what to do?

warpech:
@un-tone try the same version of Newtonsoft.Json as used here: https://github.com/Starcounter/TPCC/blob/master/TPCC/packages.config#L4

warpech:
Now the error message is just:

```
[18:04:33][restore] 'NUnit' already has a dependency defined for 'NETStandard.Library'.
[18:04:33][restore] Process exited with code 1
```

warpech:
Compare `develop` with `48_enhance-catching-rules`:

screen shot 2017-07-21 at 18 07 53

screen shot 2017-07-21 at 18 07 56

un-tone:
I downgraded back nuget packages for test project. It helped with issue.
Thanks for help @warpech

mmnosek:
> @mmnosek do you know what to do?

Unfortunately, we are pretty much stuck with this problem. Having no option to use different assemblies references makes any kind of "migration project" almost impossible. So we don't have a solution.

un-tone:
@PatrykSzwer thank you! I fixed this bug

PR #137, opened 2017-07-18
Changes in Blender API

un-tone:
Here I made changes in according to the last changes in the Blender API which described in https://github.com/Starcounter/RebelsLounge/issues/147

### DIFF
--- a/src/WebsiteProvider/Api/MappingHandlers.cs
+++ b/src/WebsiteProvider/Api/MappingHandlers.cs
@@ -113,7 +113,7 @@ namespace WebsiteProvider
         private static string GetOldUri(WebMap webMap)
         {
             var token = webMap.GetMappingToken();
-            var currentMappingUrls = new HashSet<string> { webMap.Section.GetMappingUrl() };
+            var currentMappingUrls = new List<string> {webMap.Section.GetMappingUrl()};
 
             foreach (WebMap map in webMap.Section.Maps)
             {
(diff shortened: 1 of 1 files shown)


un-tone:
@PatrykSzwer please review this PR. These changes is required for running the app in last Starcounter RC version.

src/WebsiteProvider/Api/MappingHandlers.cs
```diff
@@ -116,16 +113,15 @@ private void RegisterEmptyHandler(string uri)
         private static string GetOldUri(WebMap webMap)
         {
             var token = webMap.GetMappingToken();
-            var actualMappings = Blender.ListAll().Where(x => x.Key == token).SelectMany(x => x.Value);
-            var newMappings = new List<string> { webMap.Section.GetMappingUrl() };
+            var currentMappingUrls = new HashSet<string> { webMap.Section.GetMappingUrl() };
```
PatrykSzwer: Why did you used HashSet here? As far as I know (and I can be wrong) List is faster than hashSet in case of adding, but slower in case of looking for matching objects. And since below that we're using Add twice in a foreach, and once Contains method, isn't better to use just simple List? Again, I'm seriously asking, not suggesting new solution.

un-tone: Fixed it. You are right, here it not so justified.
FYI, I found a good explanation here https://stackoverflow.com/questions/150750/hashset-vs-list-performance

Issue #122, opened 2017-06-26
Support groups of surfaces


warpech:
Right now WebsiteEditor/WebsiteProvider support "surfaces" as the topmost concept. Depending on the specificity, surfaces may correspond to real world's:

- web pages
- websites
- fragments of web pages

Problem is that when supporting multiple systems in the same database, there is a need to group related surfaces. For example, I might have few groups of surfaces:

- Website (public)
- Extranet (for customers)
- Backoffice (for employees)

In other case, my system can be used to run dozens of websites:

- Website 1 (consists of 5 surfaces)
- Website 2 (consists of 3 surfaces)
- Website 3 (consists of 4 surfaces)
- Website 4 (consists of 1 surface)
- Admin panel for my surfaces (consists of 3 surfaces)

This clearly shows the need to group surfaces in the UI. Also, a group will have its own identity, allowing to blend additional content from other apps on the group editing level in WebsiteEditor.

How to name this functionality? "Websites", "Surface groups"?

This is a requirement for a customer project.

cc @tomalec, @miyconst, @un-tone, @PatrykSzwer

PatrykSzwer:
Since #138 is done, closing.

PR #138, opened 2017-07-18
122 groups of surfaces

PatrykSzwer:
For #122. not mergable yet. See questions below.

I'm still trying to do some styling on that input for name (I hit some bootstrap overwriting-kinda issue) and thinking about adding some button for modal window with input for renaming. What do You think about it? Also should I left functionality like it is now, with immediate transactions or should go same way like in tab views, with Save button for one transaction commit etc.? 

NOTE: To build this branch You will need newest version of Simplified master branch library.


PatrykSzwer:
It's ready to be reviewed. All needed functionalities are there.

PR #142, opened 2017-07-18
Adaptation to new pattern for session and sessionstorage

chrhol:
This PR should be seen as a sample of the changes for sessions and storing viewmodels on session. A starcounter version with the changes in branch `feature/v2.3/78-139-4017` in `Starcounter/Level1` repo are needed (since that branch is not merged yet).

Included in these changes are also the changes in branch `blender-api-upgrade` to have the app build correcctly.

Implements: https://github.com/Starcounter/Home/issues/139

Before tomorrow this must be tested with a custom build: https://teamcity.starcounter.org/viewLog.html?buildId=27821&buildTypeId=Starcounter_V23NightlyWindows&tab=artifacts

### DIFF
--- a/src/WebsiteEditor/Api/CommitHooks.cs
+++ b/src/WebsiteEditor/Api/CommitHooks.cs
@@ -20,12 +20,22 @@ namespace WebsiteEditor
 
         protected void RefreshSignInState()
         {
+            if (Session.Current == null) 
+            {
+                return;
+            }
+
             Session.ScheduleTask(Session.Current.SessionId, (session, id) =>
             {
-                var page = Session.Current.Data as MasterPage;
+                if (session == null) // Session have timed out after the task was scheduled 
+                {
+                    return;
+                }
+
+                var page = session.Store[nameof(MasterPage)] as MasterPage;
 
                 page?.Transaction.Scope(() => page.RefreshCurrentPage());
-                Session.Current.CalculatePatchAndPushOnWebSocket();
+                session.CalculatePatchAndPushOnWebSocket();
             });
         }
     }
--- a/src/WebsiteEditor/Api/MainHandlers.cs
+++ b/src/WebsiteEditor/Api/MainHandlers.cs
@@ -131,21 +131,17 @@ namespace WebsiteEditor
 
         protected MasterPage GetMasterPageFromSession()
         {
-            if (Session.Current == null)
-            {
-                Session.Current = new Session(SessionOptions.PatchVersioning);
-            }
-
-            MasterPage master = Session.Current.Data as MasterPage;
+            MasterPage master = Session.Ensure().Store[nameof(MasterPage)] as MasterPage;
 
             if (master == null)
             {
                 master = new MasterPage();
-                Session.Current.Data = master;
+                Session.Current.Store[nameof(MasterPage)] = master;
             }
-
+ 
             return master;
         }
+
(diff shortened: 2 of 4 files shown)


PatrykSzwer:
Code itself works perfectly!

src/WebsiteEditor/Api/MainHandlers.cs
```diff
@@ -131,21 +131,17 @@ protected void RegisterPartials()
 
         protected MasterPage GetMasterPageFromSession()
         {
-            if (Session.Current == null)
-            {
-                Session.Current = new Session(SessionOptions.PatchVersioning);
-            }
-
-            MasterPage master = Session.Current.Data as MasterPage;
+            MasterPage master = Session.Ensure().Store[nameof(MasterPage)] as MasterPage;
```
chrhol: It's a shortcut to make sure that a session is created and set as current. It will always return a session, and create one if `Session.Current` is `null`

Basically it's the same as doing:

```csharp
if (Session.Current == null) {
    Session.Current = new Session(SessionOptions.PatchVersioning); 
}
// Use Session.Current
```

un-tone:
Last SC build broke WebsiteEditor, most of the tests failed. Something in old Session API was changed?
cc @chrhol @warpech

chrhol:
Sorry @un-tone, we should have clarified this. It's because of https://github.com/Starcounter/starcounter-include/issues/34, this bug gets visible now that we have namespaces on root-level. 

Short term, upgrading to Palindrom 3.0.0 fixes this problem (or at least make replacing the root viewmodel work), which should be done soon in Level1 (I guess). And then of course fix `starcounter-include` properly.

src/WebsiteEditor/Api/CommitHooks.cs
```diff
@@ -20,12 +20,22 @@ public void Register()
...
+                return;
+            }
+
             Session.ScheduleTask(Session.Current.SessionId, (session, id) =>
             {
-                var page = Session.Current.Data as MasterPage;
+                if (session == null) // Session have timed out after the task was scheduled 
+                {
+                    return;
+                }
```
warpech: Is this always needed in `Session.ScheduleTask`? Why would `Session.ScheduleTask` trigger a callback for a timed out session?

chrhol: Yes, currently this is how the task is implemented. I think the argument was that if you keep track of sessions you have created, you should have an option to clean up. So if the session have timed out (or been disposed for some other reason), the parameter `session` will be `null`, but the `sessionId` parameter will still contain the id.

chrhol:
> The tests are failing

Aha, so some other problem as well. This is my last day working before vacation, so if it's anything I need to look at we need to do it soon.

warpech:
Made another build that includes Palindrom 3.0.0 and indeed tests are now passing!!!

I think this is good to merge. @un-tone if you agree, please merge.

src/WebsiteEditor/Api/CommitHooks.cs
warpech: Thanks for the explanation. I have added it to https://github.com/Starcounter/Docs/issues/84

warpech: Anyway, I think that more explicit API for session cleanup would be better. I can imagine myself writing a handler that does not check if `session` equals null.

un-tone:
Yes, looks good. Merging...

Issue #132, opened 2017-07-11
Post review for PR #129


warpech:
Regarding the line:

```
Handle.GET("/websiteeditor", () => Self.GET("/WebsiteEditor/surfaceGroups"));
```

Please use lowercase URIs: https://starcounter.gitbooks.io/guidelines/content/uris.html

------

Regarding the file `src/WebsiteEditor/wwwroot/WebsiteEditor/css/websiteeditor.css`:

In BEM, there cannot be Element classes like:

```
.website-surface-groups__button
.website-surface-groups__add-button
.website-surface-groups__plus
```

without a Block class `website-surface-groups`.

Better solution is to have two Blocks here:

```
.website-surface-groups-add
.website-surface-groups-add__icon
.website-surface-groups-link
```

PatrykSzwer:
I'll include those changes in next PR which will combine pinning rules with blending points tabs.

PatrykSzwer:
Classes are already changed (and will be change even more in #138). Regarding the URIs in handlers, take a look on commit: 22a9c4ac4c1e11828446e9484dedd44d1ef686f8 
Those will be merged in PR: #138

PatrykSzwer:
Changes are merged, please take a look on them and if everything looks ok, please close this issue. Otherwise I'll create separate PR.

warpech:
Still there are two Elements without Block, but they come from an earlier change:

- `.website-top-menu-item__surface-name`
- `.website-top-menu-item__link`

PatrykSzwer:
I've created PR for that #146

PatrykSzwer:
Final touches done in #146. Closing.

PR #146, opened 2017-07-25
Fixes due to post review.

PatrykSzwer:
#132


src/WebsiteEditor/wwwroot/WebsiteEditor/css/websiteeditor.css
```diff
@@ -22,30 +22,30 @@ body {
...
-    padding: 0px;
-    margin: 0px;
+    padding: 0;
+    margin: 0;
     overflow: auto;
 }
 
-.website-top-menu__item {
+.website-nav-element {
     float: left;
```
warpech: `float: left` is not needed when you are in `display: flex`

PatrykSzwer: Fixed.

PR #155, opened 2017-07-31
Fix slots, add declarative-shadow-dom if needed

Mihaiii:
For https://github.com/Starcounter/RebelsLounge/issues/152 .

The first commit is not really for merge at this moment. 
The PR is more so we can start a discussion about what real solution could we implement.


src/WebsiteProvider/wwwroot/WebsiteProvider/viewmodels/SurfacePage.html
```diff
@@ -5,6 +5,6 @@
         }
     </style>
     <template is="dom-bind">
-        <starcounter-include class="website-surface" slot="WebsiteProvider/Surface/{{model.WebTemplatePage.Name}}" view-model="{{model.WebTemplatePage}}"></starcounter-include>
```
PatrykSzwer: First of all I would not delete entire slot, but just left part to `/Surface`, like `slot="WebsiteProvider/Surface`, add declarative-shadow-dom template and see how is it gonna behave. 

I thinking about actually ridding of those views in websiteProvider, since this app is more like an engine not app with UI etc. itself. @warpech Do You think It would be possible to remove views and left functionality untouched?

src/WebsiteProvider/wwwroot/WebsiteProvider/viewmodels/WebsiteSection.html
```diff
@@ -2,7 +2,7 @@
 
 <template>
     <template is="dom-bind">
-        <starcounter-include slot="websiteeditor/Section/{{model.Name}}/PinContent" view-model="{{model.PinContent}}"></starcounter-include>
-        <starcounter-include slot="websiteeditor/Section/{{model.Name}}/MainContent" view-model="{{model.MainContent}}"></starcounter-include>
```
PatrykSzwer: Same as above.

src/WebsiteProvider/wwwroot/WebsiteProvider/viewmodels/SurfacePage.html
Mihaiii: In that case there will be multiple slots named `slot="WebsiteProvider/Surface`. Could that cause problems?

PatrykSzwer: Thats why I'm talking about checking it and seeing if It's gonna work, because I'm not 100% sure.

Mihaiii: I don't know how to test it **right** in this case.
Just by running it and looking around, it looks good.

PatrykSzwer: Maybe @un-tone should be the final voice about this, because He's responsible for WebsiteProvider. Let's not mess around and break something in his app (without his knowledge 😃)

Mihaiii: Ping @un-tone .

un-tone: I don't see any reason to make it just `slot="WebsiteProvider/Surface`. Why you can't do it in the same way as you did for other apps?

> I don't know how to test it right in this case.

I think you can use CompositionEditor to do some testing (if you already didn't use it)

Mihaiii: > Why you can't do it in the same way as you did for other apps?
Because the template is `declarative-shadow-dom` and in this case it would also have to be `dom-bind`. Please correct me if I'm wrong.

un-tone: TBH I'm not confident in this issue. I suppose we should ask @warpech, could you please help with it?

Mihaiii: Uh, I'll just make a new commit and include the `declarative-shadow-dom` inside the `dom-bind`, not outside of it, like it's the normal pattern. If it worked with the AppIcon (as it is in the KitchenSink), it should work it here also.

The only problem is that I'm still not sure how to test it right.

Mihaiii: Please disregard my previous comment. Doesn't work.

un-tone: I see that this `SurfacePage.html` is not needed and not used as expected anymore. I can remove it and currently investigating about `WebsiteSection.html`

un-tone: Please check PR https://github.com/StarcounterApps/Website/pull/161

warpech: > Because the template is `declarative-shadow-dom` and in this case it would also have to be `dom-bind`. Please correct me if I'm wrong.
> LE: Or can it be both?!

No, it cannot be both. Technically it doesn't work and also it is against the point of separating the concerns between the content and the layout.

Mihaiii:
@PatrykSzwer Please have another look.

Issue #139, opened 2017-07-18
Error when saving a pinning rule with same properties


un-tone:
### Steps to reproduce

1. Start WebsiteEditor, WebsiteProvider and SignIn from `develop` branch.
2. Go to `http://localhost:8080/websiteeditor/resetdata` to reset WebsiteEditor data.
3. Go to "Pinning rules" page for Default surface. There is one pinning rule.
4. Add to the rules another one with the same values and press "Save".

### Expected result

The pinning rules saved without errors

### Actual result

The WebsiteProvider throws an error with the following message: `The following URI is already added to blender: "/signin/user" on "websiteeditor%B%C"`.

##### Involved apps' version

 - StarcounterApps/Website@75b3706947a0d5a980830592294adb9953319fb6
 
### Screenshot
![image](https://user-images.githubusercontent.com/1625378/28315656-15b1b64c-6bc8-11e7-8c85-10c40caa49a3.png)

un-tone:
Created PR https://github.com/StarcounterApps/Website/pull/158

un-tone:
Closing as done

PR #158, opened 2017-08-02
Fixed error when mapping duplicated pinning rules

un-tone:
For https://github.com/StarcounterApps/Website/issues/139.

With this fix user can save duplicated pinning rules, but it will not affect on its mapping. It is supposed that in future WebsiteEditor will have user notification feature implemented and WebsiterEditor will not allow to save duplicated pinning rules. But still in such case this fix will be required.

### DIFF
--- a/src/WebsiteProvider/Api/PinningHandlers.cs
+++ b/src/WebsiteProvider/Api/PinningHandlers.cs
@@ -14,7 +14,7 @@ namespace WebsiteProvider.Api
     {
         private static PinningHandlers instance;
         private bool isRegistered;
-        private PinningRulesMappingCollection mappings = new PinningRulesMappingCollection();
+        private readonly PinningRulesMappingCollection mappings = new PinningRulesMappingCollection();
 
         private PinningHandlers()
         {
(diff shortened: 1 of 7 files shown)


src/WebsiteProvider/Api/PinningHandlers.cs
```diff
@@ -282,5 +280,44 @@ private class PinningRuleMappingInfo
             public string Token { get; set; }
             public string TokenBase { get; set; }
         }
+
+        private class PinningRulesMappingCollection
```
PatrykSzwer: I would suggest moving PinningRulesMappingCollection class to separate file, even if It's used only here.

un-tone: Done.

un-tone:
@PatrykSzwer I can't re-produce your bug. Can you provide more info and detailed steps for it?

un-tone:
Thanks! I reproduced the first bug with deleting.
But the last one is not reproducible for me.

PatrykSzwer:
A little bit later I'll try to reproduce it and describe you all steps.

un-tone:
I think the bug with deleting should be fixed in the new separate issue.

Mihaiii:
The "delete bug" seems the one from here: https://github.com/StarcounterApps/CssVariablesManager/issues/25 .
Please read the comments in that issue. Hope that will save you some time as it doesn't seem related to Website app (if it's the same bug) 👍

PatrykSzwer:
I agree with separating the issues.

PatrykSzwer:
I can't reproduce this issue that cause exception. Maybe It was also related to the previous issue with triple entries. I hope It's not going to show up again.

PR #160, opened 2017-08-03
Removed juicy-composition.

PatrykSzwer:
Related to https://github.com/Starcounter/RebelsLounge/issues/156
It's not mergable yet!

I removed juicy-composition from default-surface, but TBH I don't know if I did it in proper way, specially because I can't test it yet, so please take a look. 
@tomalec Could You point me out what I have to do to be able to test it properly?


PatrykSzwer:
My question was more like, how should I update my starcounter-include to 3.0 version with that feature? Just use bower to update it to branch version?

tomalec:
Youi can also use custom build of Starcounter http://teamcity.starcounter.org/viewLog.html?buildId=28285&buildTypeId=Starcounter_V23NightlyWindows&tab=artifacts

PatrykSzwer:
For some reason with that custom build, entire website is not working/showing up (same with or without juicy-composition)

warpech:
Now there's a newer build: http://teamcity.starcounter.org/viewLog.html?buildId=28938&buildTypeId=Starcounter_V23NightlyWindows&tab=artifacts

Perhaps you don't see the elements because of some slot attributes do not have the corresponding slot elements. See: https://github.com/Starcounter/RebelsLounge/issues/160. This is fixed in https://github.com/StarcounterApps/Website/pull/155.

PatrykSzwer:
I've tested this with newest SC build (v. 2.3.1.6924) and looks like everything works fine. Please review and confirm, but I think It can be merged.

PatrykSzwer:
This should be merged immediately  after https://github.com/Starcounter/level1/pull/4240 is merged.

tomalec:
@PatrykSzwer I think we should remove a new version.
Do you think we should bump major or just minor?

Issue #97, opened 2017-05-18
It is not possible to blend main response and pinned content


warpech:
Even though the main response and the pinned content are inside the same named blending point in the Website app, they are not blendable, because they show up in different `starcounter-includes`.

To make everything blendable, this view should have a single `starcounter-include` in it: https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteProvider/wwwroot/WebsiteProvider/viewmodels/WebsiteSection.html#L6

Discuss this after PR #70 is merged.

warpech:
I think that #97 should not be worked on right now, because it is a lot of effort and little benefit right now.

un-tone:
Created PR https://github.com/StarcounterApps/Website/pull/161 for this issue

un-tone:
Closing as done

Issue #154, opened 2017-07-28
Broken integration with Content app


Mihaiii:
### Steps to reproduce
1. Run in debugging mode [Content app](https://github.com/StarcounterApps/Content) at https://github.com/StarcounterApps/Content/pull/92/commits/89c69d359f48dda2d4beb3768c4e7dcf71ef9c2d .
2. Run Website at  77cbcd865ab1891e26b265426bf954123e57ee35 .
3. Run Sign In at https://github.com/StarcounterApps/SignIn/commit/bb24f80f31f59a80b925d4b382edaa133f0d9cb9
4. Put a breakpoint [here](https://github.com/StarcounterApps/Content/pull/92/commits/89c69d359f48dda2d4beb3768c4e7dcf71ef9c2d#diff-ccd4cf3920c56e95dcc937e7fa590631L17) .
5. Go to `http://localhost:8080/resetdata` -> `http://localhost:8080/websiteeditor` -> "DefaultSurface" -> "Blending points" and replace for "TopBar" the "Pin URI" from `/signin/user` to `/content/dynamic/something`.

### Expected result
Hit the breakpoint.

### Actual result
It doesn't hit the breakpoint.

### More Info
It works fine for `/signin/user` or `/signin/profile`.
**Update: doesn't work with `/signin/signinuser?something`**

##### Starcounter version 
SC 2.3.1.6852

@alemoi , @un-tone Can you help with this one?

un-tone:
I'll check it

un-tone:
I see the following case:
 - Content app has a handler for `/content/dynamic/{?}` URI which accept `/content/dynamic/something` requests.
 - Blender mapped this URI correctly:
```
Blender.ListForSpecificUri("/websiteprovider/blender/surface/y/point/z").Select(x => x.Uri)
    [0]: "/content/dynamic/something"
    [1]: "/websiteprovider/blender/surface/y/point/z"
```
Website calls `/websiteprovider/blender/surface/y/point/z` URI to get content for "TopBar".

-----

Blender expects `/content/dynamic/{?}` to be mapped -- it cannot blend just `/content/dynamic/something` since this URI is not registered with Handler.
There is a problem for Website app, it definitely should be able to map such URIs as expected.

@alemoi is it possible do something with it on the Blender side? If no, I suppose I'll have to wrap such custom URIs callings into another dynamically created one. I wouldn't do so since it will add complexity to pinning rules mapping logic.

alemoi:
[...]

PR #161, opened 2017-08-03
Improved blending of main and pin contents

un-tone:
Implemented https://github.com/StarcounterApps/Website/issues/97. Also I guess it can be related to https://github.com/StarcounterApps/Website/pull/155.

Currently it is a POC. The goal is ridding of HTML files in WebsiteProvider and improving blending of main response and pinning content.

I don't fully understand what some pieces of code is doing and I'd appreciate if @warpech explain me the following:

1. https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteProvider/Api/ContentHandlers.cs#L170 - what exactly this code checks:
```csharp
sectionJson.MainContent.LastJson != json
```
What is the reason to have `LastJson`

2. https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteProvider/Api/ContentHandlers.cs#L162 - what is the case for this code:
```csharp
json is SurfacePage
```

The problem with these pieces of code that with my changes I will have to replace `content.Sections[section.Name]` value with another `Json` instance but in this case the app is not working correclty. I tried to do it differently but without success. So currently I just commented it.


warpech:
> `json is SurfacePage`

I don't remember the reason for it. I have added it in [this commit](https://github.com/StarcounterApps/Website/commit/0942536571fd0aec2edabac4109f59462003c7c8), by refactoring what was previously [this line](https://github.com/StarcounterApps/Website/blob/063474946a83046e4ef7f5a4d5a40eea634508a5/src/WebsiteProvider/Api/ContentHandlers.cs#L40). My best guess is that it is just an artifact from the times when WebsiteEditor and WebsiteProvider were one app.

I put a breakpoint in it and ran all the tests and it is never hit.

warpech:
> `sectionJson.MainContent.LastJson != json`

This line is not insignificant. But as it turns out, there is no test for the differences. 

Remembering the `lastJson` allows to reuse the previous `MainContent` instead of rewriting it. Because of this we get smaller patches. Sadly, this seems to be the only benefit. And the patch difference is smaller than I would expect. I guess it should be improved or removed.

## How to reproduce the difference

Consider clicking on "Blending points" when being in http://localhost:8080/WebsiteEditor/surface/NW/catchingrules

- The patch with `/*sectionJson.MainContent.LastJson = json*/`: https://gist.github.com/warpech/d6418347958eb3cbff5af607c6661e53
- The patch with `sectionJson.MainContent.LastJson = json`: https://gist.github.com/warpech/70395a8abfffcdc71bcc572bd56fc308

Use a JSON comparison tool to find the difference: http://www.jsondiff.com/

I hope this helps to get your head around it.

warpech:
First of all, it is amazing initiative to simplify this code!

Sadly, I've found some problem:

## Steps to reproduce

1. Start WebsiteEditor, WebsiteProvider (this PR branch)
2. Go to `http://localhost:8080/websiteeditor/resetdata` to reset WebsiteEditor data.
3. Start SignIn, People, Images (develop branch)
4. Go to `http://localhost:8080/WebsiteEditor/surfacegroups`
5. Go to DefaultSurface > BlendingPoints
6. Move the pinning rule `/signin/signinuser` from TopBar to Main
7. Save
8. Go to `http://localhost:8080/people/persons`
9. Create some person or click on "admin"
9. Change the URL from `http://localhost:8080/people/partials/persons/<id>` to `http://localhost:8080/people/persons/<id>` and load that page

## Expected

A page should load that has 3 apps in the main area (People, Images, SignIn)

## Actual

The page only has People and SignIn. Images looks as unmapped.

## Screenshot

2017-08-04 19_23_17-people

un-tone:
> `json is SurfacePage`

So I removed this code.

Thanks for the explanation @warpech and for the found bug.
I was sure that there are some bugs/issues still :smile:. I will investigate later what I can do with code related to `LastJson`.

-----

@alemoi is there any way to blend already blended JSON instance in this way:

```csharp
var newJson = Self.GET<Json>(uri, () => externalJson);
```

Here `externalJson` is a JSON handled by WebsiteProvider (this JSON in our case is a People app response blended with an Image app response). In the presented code `uri` is mapped with other URIs. The code works well if calling `uri` returns null or `externalJson` is not blended (has no siblings). Otherwise it seems like Blender misses siblings from `externalJson`.

If it is not possible to "reblend" JSONs, maybe we could add such feature?

alemoi:
> Here externalJson is a JSON handled by WebsiteProvider (this JSON in our case is a People app response blended with an Image app response). In the presented code uri is mapped with other URIs. The code works well if calling uri returns null or externalJson is not blended (has no siblings). Otherwise it seems like Blender misses siblings from externalJson.

It could be a problem with JSON response merger. If you can give me the exact case to repro, I will check it :)

un-tone:
Sure, we have it - Website app, `improve-blending-wp` branch, this case https://github.com/StarcounterApps/Website/pull/161#issuecomment-320306104.
You can debug at this line [ContentHandlers.cs#L170](https://github.com/StarcounterApps/Website/blob/improve-blending-wp/src/WebsiteProvider/Api/ContentHandlers.cs#L170)

alemoi:
When I follow your reproduction steps, right when I call `http://localhost:8080/websiteeditor/resetdata` I get the following exception:
```
System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.ThrowHelper.ThrowKeyNotFoundException()
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at WebsiteProvider.Api.PinningHandlers.UnmapPinningRule(PinningRuleMappingInfo info, Nullable`1 certainMapId) in c:\github\Website\src\WebsiteProvider\Api\PinningHandlers.cs:line 240
   at WebsiteProvider.Api.PinningHandlers.UnmapSurface(UInt64 templateId) in c:\github\Website\src\WebsiteProvider\Api\PinningHandlers.cs:line 212
   at WebsiteProvider.Api.CommitHooks.<Register>b__4_4(Object s, UInt64 id) in c:\github\Website\src\WebsiteProvider\Api\CommitHooks.cs:line 40
   at Starcounter.RuntimeInstalledHookDelegate`1.Invoke(Object triggeringObject) in C:\github\level1\src\Starcounter\Hooks\RuntimeInstalledHookDelegate.cs:line 37
   at Starcounter.InvokableHook.InvokeAllWithKey(HookKey key, Object instance) in C:\github\level1\src\Starcounter\Hooks\InvokableHook.cs:line 73
   at Starcounter.InvokableHook.InvokeDelete(HookKey key, UInt64 objectID) in C:\github\level1\src\Starcounter\Hooks\InvokableHook.cs:line 61
   at Starcounter.Internal.TransactionManager.InvokeHooks() in C:\github\level1\src\Starcounter\TransactionManager.cs:line 812
   at Starcounter.Internal.TransactionManager.Commit(UInt64 handle, Int32 free) in C:\github\level1\src\Starcounter\TransactionManager.cs:line 704
   at Starcounter.Db.TransactAsync(Action action, UInt32 flags, TransactOptions opts) in C:\github\level1\src\Starcounter\Db.cs:line 363
   at Starcounter.Db.Transact(Action action, UInt32 flags, TransactOptions opts) in C:\github\level1\src\Starcounter\Db.cs:line 467
   at Starcounter.Db.Transact(Action action, Int32 maxRetries) in C:\github\level1\src\Starcounter\Db.cs:line 265
   at WebsiteEditor.Helpers.DataHelper.ClearData() in c:\github\Website\src\WebsiteEditor\Helpers\DataHelper.cs:line 10
   at WebsiteEditor.Api.MainHandlers.<>c.<Register>b__0_2() in c:\github\Website\src\WebsiteEditor\Api\MainHandlers.cs:line 38
   at Starcounter.Rest.UserHandlerInfo.RunUserDelegate(Request req, IntPtr methodSpaceUriSpaceOnStack, IntPtr parametersInfoOnStack) in C:\github\level1\src\Starcounter.Rest\UriHandlersManager.cs:line 231
   at Starcounter.Internal.Web.AppRestServer.RunDelegateAndProcessResponse(IntPtr methodSpaceUriSpaceOnStack, IntPtr parametersInfoOnStack, Request req) in C:\github\level1\src\Starcounter.Apps.JsonPatch\AppRestServer.cs:line 96
   at Starcounter.Internal.AppsBootstrapper.ProcessExternalRequest(Request req) in C:\github\level1\src\Starcounter.Apps.JsonPatch\AppsBootstrapper.cs:line 773
HResult=-2146232969

More information about this error may be available in the server error log.
```

un-tone:
@alemoi what SC version do you use? It should be 2.3.1.6886. If you use the same version, could you please try again on clean DB?

alemoi:
@un-tone yesterday I pushed it to Starcounter 2.3 and develop so its going to be there. I will try with a fresh db.

un-tone:
@alemoi do you have any updates?

alemoi:
Will check this one today. Didn't have time before, sorry.

alemoi:
@un-tone as promised (though not same day :D) but I tried to repro exactly as written in https://github.com/StarcounterApps/Website/pull/161#issuecomment-320306104
with LATEST 2.3 and all worked fine (both URIs showed all 3 apps presence). So I think I might have fixed this one automatically with this: https://github.com/StarcounterApps/Website/issues/154

@un-tone can you please try it with latest v2.3? Thanks a lot!

un-tone:
It seems that the issue is not fixed. It works in the same way in the latest 2.3.1.6981 SC version.

This is how it looks with `http://localhost:8080/people/persons/id` URL
image

And this is with `http://localhost:8080/people/partials/persons/id` URL
image

alemoi:
@un-tone thanks for providing more info. I checked the code and its not related to Blender, as I see. Its rather a `RunResponseFilters` that does not properly work out with existing response with siblings. I am not really familiar with the code, so will have to ask @chrhol tomorrow for some help.

alemoi:
@warpech thanks for the reproduction description scenario. The last step seems a bit wrong though, the URIs are the opposite that are working and non-working. Please add Admin user creation step in SignIn next time :) Thanks :)

alemoi:
After debugging with @chrhol he has promised to check the problems there. Its related to how siblings are merged together (Images sibling gets removed by some reason).

alemoi:
Just corrected the reproduction scenario: https://github.com/StarcounterApps/Website/pull/161#issuecomment-320306104

chrhol:
Sorry for the delay in getting into issue.

I think the following is something we really need to discuss, and that is the current problem:  

> @alemoi is there any way to blend already blended JSON instance in this way:
> ```csharp
> var newJson = Self.GET<Json>(uri, () => externalJson);
> ```

Doing another round of blending and using an already blended viewmodel currently behaves in an undefined way, and I'm not really sure how the logic behind it should look. What should the outcome be?
Sometimes everything should be combined and sometimes some partial viewmodels should be removed. The problem is that there is no good way of knowing this. 

What is the reason for doing it in this way? To allow more than one response for a website-section? Then maybe we should have some manual way to combine two different blending results (kind of like `.MergeJson(...)` but more explicit).

@tomalec: I think we should bring this up on the next blending tf meeting and find some good patterns and what should be allowed or not.

un-tone:
Currently I partially addressed the issue with already blended JSON by wrapping it with another JSON.
FYI:

```csharp
Self.GET<Json>(uri, () => section.Default
    ? (json.Parent ?? new ContentWrapperPage {Content = json})
    : new Json());
```
If not to check `json.Parent` I get the following error: `System.Exception: Cannot change parent of objects in Typed JSON trees`

-----

@PatrykSzwer please review this PR. FMPOV it is ready to be merged.

-----

> What is the reason for doing it in this way?

Using this way I will get rid of wrapping JSON main content in the WebsiteProvider. We assume that this app should not contain any HTML files except templates. But now I think it is not so important at this stage.

> Then maybe we should have some manual way to combine two different blending results (kind of like .MergeJson(...) but more explicit).

Yeah, such feature would be very useful.

un-tone:
Thanks @PatrykSzwer

un-tone:
The implementation is changed here #205

Issue #123, opened 2017-06-26
Implement Authorization library


warpech:
This app needs to implement the Authorization library to allow managing what users have rights to administer surfaces, etc.

warpech:
Reference implementations:

- https://github.com/StarcounterApps/Images
- https://github.com/StarcounterApps/Calendar

@un-tone will add comments to the Images app and then try to write the article for the Cookbook. The article will be similar to https://docs.starcounter.io/guides/web-apps/acceptance-testing-with-selenium/ (btw. that article should be moved to Cookbook).

PatrykSzwer:
Auth library was implemented in #162 . Closing.

un-tone:
Unfortunately it seems like these changes breaks refreshing page after logging in:
![image](https://user-images.githubusercontent.com/1625378/29034603-7d4dcbea-7ba1-11e7-8cf3-146522d4356f.png)
@PatrykSzwer could you please check it?

PatrykSzwer:
I'll check it and create new PR for this.

PR #162, opened 2017-08-04
123 auth library

PatrykSzwer:
For #123 

Authorization library is initially implemented. All pages are loaded trough library/MasterPageMiddleware, all handlers for pages in mainHandlers were removed. Right now there is one rule added, to show surfaceGroupsPage only for user named "admin".

Now only thing left is to discuss which rules should we add initially to WebsiteEditor, and eventual UI to edit authorization rules (I would suggest to create separate issue for that). Since this functionality was requested by Client maybe we should add some case related to his needs? @warpech What do You think?


src/WebsiteEditor/Api/MainHandlers.cs
```diff
@@ -40,60 +40,6 @@ public void Register()
...
-            });
-
-            Handle.GET("/websiteeditor/surface/{?}/catchingrules", (string key) =>
-            {
-                return Db.Scope<MasterPage>(() =>
-                {
-                    MasterPage master = this.GetMasterPageFromSession();
-                    master.ShowNavigation = true;
-                    master.Surface.Data = Db.SQL<WebTemplate>("SELECT t FROM Simplified.Ring6.WebTemplate t WHERE t.Key = ?", key).First();
-                    SetMasterCurrentPage(master, "/websiteeditor/partials/catchingrules");
```
un-tone: `SetMasterCurrentPage` and `GetMasterPageFromSession` methods are not used anymore. Please remove it.

src/WebsiteEditor/ViewModels/MasterPage.json.cs
```diff
@@ -24,7 +24,7 @@ public void RefreshCurrentPage()
 
             SystemUser user = SystemUser.GetCurrentSystemUser();
 
-            if (SystemUser.IsMemberOfGroup(user, allowedSystemUserGroup))
+            if (SystemUser.IsMemberOfGroup(user, AllowedSystemUserGroup))
```
un-tone: These checks looks redundant since we currently use Authorization lib. I think we should get rid of theses checks in RefreshCurrentPage method.

src/WebsiteEditor/ViewModels/TestPage.json.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace WebsiteEditor.ViewModels
+{
+    [Database]
+    public class TestObject : Something
+    {
+        public bool IsThisWorking { get; set; }
+    }
+
+    [PartialUrl("/websiteeditor/partials/test")]
```
PatrykSzwer: Oh, those are some leftovers after me trying auth library. No idea why those left, I'll delete everything thats not needed.

src/WebsiteEditor/ViewModels/MasterPage.json.cs
PatrykSzwer: I agree. That's good idea to consider this PR as replacing old authorization methods by authorization library. Eventual expanding of functionalities will be done in separated issues/PRs.

src/WebsiteEditor/Api/Authorization/AuthEnforcementProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            get
+            {
+                if (_instance == null)
+                {
+                    lock (ThreadSync)
+                    {
+                        if (_instance == null)
+                        {
+                            _instance = InitializeAuthEnforcement();
```
PatrykSzwer: I've added comment for that in this PR: https://github.com/StarcounterApps/Website/pull/169

> Singleton thread save pattern with usage of Double-Check Locking. 
Safety for situation where two threads ask about an instance and without lock could get two of them.

un-tone: @PatrykSzwer that comments doesn't explain the case why this locking is required. It just described the pattern, and I think it is redundant.

PatrykSzwer: Fixed.

Issue #163, opened 2017-08-04
Wrong generated catching rule


warpech:
What is 

```
/websiteprovider/surfaces
```

Should be:

```
/websiteeditor/surfacegroups
```

in https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteEditor/Helpers/DataHelper.cs#L107

----

I assume that `/websiteprovider/help?topic=surfaces` in that file should also be changed to `websiteeditor`.

PatrykSzwer:
Created PR for that: #165

PatrykSzwer:
Closing as done.

PR #165, opened 2017-08-07
Fixes on catching rules in dataHelper.cs. #163

PatrykSzwer:
For #163

### DIFF
--- a/src/WebsiteEditor/Helpers/DataHelper.cs
+++ b/src/WebsiteEditor/Helpers/DataHelper.cs
@@ -104,10 +104,10 @@ namespace WebsiteEditor.Helpers
             var templatesUrl = new WebUrl
             {
                 Template = surface,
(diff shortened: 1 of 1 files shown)


warpech:
Code changes look good but something is still wrong with the pinning rule.

## Expected

The left sidebar column on `/websiteeditor/surfacegroups` should render the help topic "What is a surface?"



## Actual

The left sidebar column does not show the help topic.



2017-08-07 17_38_49-website running - microsoft visual studio

PatrykSzwer:
It's related with https://github.com/StarcounterApps/Website/issues/154. Currently there is a problem with parameterized URIs. If You try with any uri without parameters, then It should work. Since It's not part of this issue please confirm if It's mergable.

un-tone:
FYI, the issue with rendering of the "What is a surface?" area was fixed

PR #168, opened 2017-08-08
Updated Blender API using

un-tone:
This changes is for last RC version 2.3.1.6903


warpech:
Please check the last commit in the `develop` branch by @alemoi.

alemoi:
> @alemoi please follow the PR way of making code changes, so that other developers can get familiar with the proposed changes.

I now include everything in change log and linking issues in commits. The API at this stage is quite unstable, so I change it without doing "obsolete", because I would end up in maintenance.

PR #171, opened 2017-08-09
Fix on surface groups view

PatrykSzwer:
Fixed view of surface groups to be displayed correctly with new version o SC.

PR #172, opened 2017-08-09
Fix the way content is distributed for SurfaceGroupsPage.html

tomalec:
An alternative way to https://github.com/StarcounterApps/Website/pull/171


tomalec:
Changes:
 - Fixes flaky auto- slot name assignment for dynamically created elements, (but this time by..)
 - Adds explicit slot elements and attributes - makes code more self-explanatory,
 - Allows blending individual surface groups, for example displaying them in CSS Grid,
 - Hides styling related elements `<hr>` in Shadow DOM.
 - Moves local styles to the Shadow DOM,
 - Does not relly on problemativc bootstrap to style `<hr>`.

tomalec:
Sorry, I didn't know that was the intention, do you think CSS styling would be good enough replacemen, or do you really need `<hr>` element there for some reason?

PatrykSzwer:
I just think that those hr lines make it more readable and looks nice. But if You'll show comparatively good solution, I'm open for it!

tomalec:
Here is my proposition:
obraz

Howeve, I still wigth with a V0 polyfill bug.

PatrykSzwer:
This looks awesome! As soon as You deal with this polyfill bug I definitely want to try this out! Make sure that adding more than 3-4 surfaces to group won't break anything.

tomalec:
Here you have the results. it even looks more-or-less good in IE with `flex` instead of `grid`.

Fix for `::content` support will follow in `stracounter-include@3.0.x`

@warpech do you think it's worth making the fix for pre 3.0.0 version? (in juicy-composition)

PatrykSzwer:
Wow, looks really good! (and complex) But the view in Edge, FF and Chrome are all different. I'm not sure if we should do this in that way, where We have sightly different views on all browsers.

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html
```diff
@@ -2,44 +2,124 @@
...
+                grid-template-columns: repeat(3, 1fr);
+                margin-bottom: 3em;
+            }
+            .websiteeditor-hr{
+                margin: 23px 0;
+                box-sizing: content-box;
+                height: 0;
+                border-top: 1px solid #eeeeee;
+            }
+            polyfill-next-selector { content: '[slot="websiteeditor/groups"]'; }/* for polyfilled browsers*/
```
PatrykSzwer: I never saw this element before. Is there any documentation for this `polyfill-next-selector` ?

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html
```diff
@@ -2,44 +2,124 @@
...
+            ::slotted([slot="websiteeditor/add-group"]){
+                align-self: center;
+            }
+            polyfill-next-selector { content: '[slot="websiteeditor/groups"], .websiteeditor-unassigned'; }/* for polyfilled browsers*/
+            .websiteeditor-groups ::slotted(div), .websiteeditor-unassigned{
+                padding: 1em;
+                border-radius: 4px;
+                overflow: hidden;
+                background: white;
+                box-shadow: 0 1px 4px 0 var(--websiteeditor-shadow-color, hsla(0, 0%, 0%, 0.1)), 0 0px 4px 0 var(--websiteeditor-shadow-color, hsla(0, 0%, 0%, 0.1));
```
PatrykSzwer: Why are You defining `--websiteeditor-shadow-color` like that, since You have definition of it above, in host? And why even using variable here for box-shadow?

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html
```diff
@@ -2,44 +2,124 @@
...
+            }
+            polyfill-next-selector { content: '[slot="websiteeditor/groups"]:hover, .websiteeditor-unassigned:hover'; }/* for polyfilled browsers*/
+            ::slotted(div:hover), .websiteeditor-unassigned:hover {
+                box-shadow: 0 1px 4px 0 var(--websiteeditor-shadow-color, hsla(0, 0%, 0%, 0.1)), 0 0px 4px 0 var(--websiteeditor-shadow-color, hsla(0, 0%, 0%, 0.1)), 0 0 20px 0 var(--websiteeditor-shadow-color, hsla(0, 0%, 0%, 0.1));
```
PatrykSzwer: Same as above

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html
tomalec: https://github.com/webcomponents/webcomponentsjs/blob/v0/README.md#shadowcss-host-contexthost-doesnt-work-

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html
tomalec: To have it cleary visible that it's the same value.. I think it's a handy thing to be able to chenge it in single place, and to have names which may distingush conceptually from accitentaly saame values.

tomalec:
>  the view in Edge, FF and Chrome are all different

That's why the fix for `starcounter-include` is needed.

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html
warpech: What's the rationale of having exactly this one value as a variable? The background and the text color are not in variables anyway.

I think that repeating the default value in each usage instance of `var(--websiteeditor-shadow-color, hsla(0, 0%, 0%, 0.1))` is an overkill. The whole point of variables is that they are defined once. Paper elements do not provide fallback values in `var()`: https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L235

tomalec: fair point.

So would you suggest to remove vars complately, or remove defaults and make other colors use vars?

warpech:
> @warpech do you think it's worth making the fix for pre 3.0.0 version? (in juicy-composition)

No, unless you think that it is many days until 3.0.0 can be released in Starcounter?

PatrykSzwer:
I have some proposition, let's for now, reopen and merge https://github.com/StarcounterApps/Website/pull/171 as a fix for discovered issue, and let's consider this as an improvement/enchantment/update of SurfaceGroups view, since this got much further than just fixing an issue. Then I'll release fixed version, and this visual changes will come with next release. Can We do it this way?

tomalec:
@PatrykSzwer as fixes are now merged to 2.3.1.7022, should we come back to this one?

PatrykSzwer:
Actually, as I've heard there will be meeting about Interface of website this week. I think We can hold It till then, and we will see what will came up there.

tomalec:
Are there any updates on the new design?

PatrykSzwer:
Turns out that we Didn't discussed yet design itself of app, more like data model and general structure.

warpech:
@tomalec the app will be redesigned anyway, so we'd rather not merge this change unless it contains crucial changes. Is there something crucial in it?

tomalec:
Nothing that fixes functionality, but it
 - removes `<br/>` from light DOM, meaningful, functional elements (bad practise). So it will not end up copy and pasted in other apps
- showcases one way how to use `slot` attribute with `dom-repeat` to gather a group of dynamically created elements together,
- showcases how to use CSS Grid Layout,
- showcases hot to use `polyfill-next-selector`

If we have such examples somewhere else, I'm fine with disregarding this one.

warpech:
> - removes `<br/>` from light DOM, meaningful, functional elements (bad practise). So it will not end up copy and pasted in other apps

I have created an issue for that: https://github.com/StarcounterApps/Website/issues/207

- showcases one way how to use `slot` attribute with `dom-repeat` to gather a group of dynamically created elements together

Examples should be in KitchenSink. I have created an issue for that: https://github.com/StarcounterApps/KitchenSink/issues/169

> - showcases how to use CSS Grid Layout,

I have created an issue for that: https://github.com/StarcounterApps/KitchenSink/issues/170

- showcases hot to use `polyfill-next-selector`

I have created an issue for that: https://github.com/StarcounterApps/KitchenSink/issues/171

warpech:
Closing this as too complex. Thank's a lot @tomalec anyway. Please proceed with separate issues linked above.

Issue #176, opened 2017-08-15
Starcounter.DbException - ScErrReadOnlyTransaction


PatrykSzwer:
### Steps to reproduce
1. Reset data.
2. Click on DefaultSurface
(If error didn't appear, go back to SurfaceGroups and click DefaultSurface again.)
### Expected result
No errors.
### Actual result
Error with connection lost is showing up:
![image](https://user-images.githubusercontent.com/25042256/29317587-f0d3002c-81cc-11e7-96b7-534b0f6cc197.png)

##### Starcounter version 2.3.1.6924

As We already discussed with @joozek78 It can be related to authorization library and DbScopeMiddleware, but as It wasn't discovered before and It's pretty easy to find bug, It can also be related to SC version. Needs further investigation. 

cc @un-tone

un-tone:
Part of discussion is here https://github.com/StarcounterApps/Website/pull/178. I still have not found a reason of the issue. I already don't have time to do it. But I found out that this https://github.com/StarcounterApps/Website/pull/161 implementation works well without that bug. So I have to ask @chrhol to escalate that #161 issue.

un-tone:
After changes in SC done for https://github.com/StarcounterApps/Website/issues/177 the behaviour was changed a little. But it is still reproducible if you will got catching rules of SidebarSurface and uncheck "Nestable" checkbox for the rule.

un-tone:
I checked the issue in #161 with last changes. Unfortunately that changes does not fix the issue, maybe I missed something last time.

un-tone:
### Steps to reproduce

1. Run **WebsiteEditor** and **WebsiteProvider** apps from `develop` branch
1. Run **SignIn** app from `develop` branch
1. If a user is not created go to <http://localhost:8080/signin/generateadminuser>
1. Go to <http://localhost:8080/signin/signinuser> and log in
1. Go to <http://localhost:8080/websiteeditor>
1. Go to `SidebarSurface` page (click on the link)

### Expected result

The page is loaded without errors.

### Actual result

Get the error:

```
Starcounter.DbException: ScErrReadOnlyTransaction (SCERR4093): The transaction is readonly and cannot be changed to write-mode.
Version: 2.3.1.7063.
Help page: https://github.com/Starcounter/Starcounter/wiki/SCERR4093.
[...]

PR #178, opened 2017-08-15
Fixed surface page error

un-tone:
Fixed https://github.com/StarcounterApps/Website/issues/176


PatrykSzwer:
After jumping from revision to revision, and looking where this problem occured, I finally found out what caused this.

It's not Auth library, and not any other our straight code changes. It's... SidebarSurface. When You change sidebarSurface catching rule to any other than surfaceGroups, everything is gonna start working properly. 
Of course I'm not saying that definitely something is wrong with SidebarSurface itself. That can actually show another problem that is behind all of this. But We finally have some point to start looking.

PatrykSzwer:
Seems like It's related with issue about registering URI with arguments. After I've changed way of using help handler everything looks ok.

un-tone:
The error happens when using any nested surface (not only SidebarSurface) - you can move that catching URI to the HolyGrailSurface and get the same error. But the error does not happen when nesting is not set for requested pages.

I see the issue with Transaction on pages when the error happens - it is `null` in this case.
I continue investigating.

PR #182, opened 2017-08-21
Blending with Context

habib535:
https://github.com/Starcounter/RebelsLounge/issues/126


warpech:
This seems to work ok after `develop` is merged to `context`.

Issue #54, opened 2017-03-06
Support regex in Catch Uris


joozek78:
Outline
---

You want to assign a layout to whole app, but you don't know the exact URLs of this app. There is a real scenario (Sweoffshore) where you create a layout for a suite that might expand in the future and you can't predict exact URLs in those apps . Catch Uris only support wildcards for one segment, so `/App/{?}` would catch `/App/abc`, but wouldn't catch `/App/a/b/c`.

Workaround
---

Create a bunch of Catch Uris: `/App/{?}`, `/App/{?}/{?}`, and so on. The downside is that you can't refer whole group of them when creating Pinning Rule.

Proposed solution
---

Support regular expressions, not just "{?}" wildcard. That way you can cover arbitrary set of URLs

Anticipated problems
---

Conflicting rules. Those conflicts could be resolved by adding "order" field or by some logic based on specificity (like CSS does)

un-tone:
Closing as done

PR #190, opened 2017-09-07
Regex support for catching rules

un-tone:
For #54 
I'm considering this feature as experimental so I have not changed the data model. I just use prefix `\regex\` to mark regex rules. For example, use `\regex\\/WebsiteEditor\/surface\/(\w|\d)*\/general` catch URI to apply catching rule for general pages of all surfaces.

Currently we have the following priorities for catch URI matching:

1. Exact match
2. Match with wildcard (described in README)
3. Match with regex
4. Catch-all

image

This PR should not close the #54 issue since we should make additional improvements.
Here I'd want to get a feedback on my chosen approach.


PatrykSzwer:
I think, speaking about UI itself, It should be some indicator like checkbox, to say that "This URI is with REGEX", not a part or URI itself. It would be more intuitive and readable.

un-tone:
We should consider these #191 changes when doing it.

warpech:
I agree with @PatrykSzwer regarding the checkbox (like in Notepad++)

un-tone:
Please review the PR. I've updated UI and added new test.

un-tone:
Requires https://github.com/StarcounterApps/Simplified/pull/75

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/CatchingRulesPage.html
```diff
@@ -37,11 +37,15 @@
                     </div>
                     <div class="website-paper-collapse-item-text-input-section">
                         <label for="websiteCatchingRuleUri">Catch URI: </label>
-                        <input type="text" id="websiteCatchingRuleUri" value="{{catchingRule.Url$::change}}" class="website-paper-collapse-item-text-input-section__input"/>
+                        <input type="text" id="websiteCatchingRuleUri" value="{{catchingRule.Url$::change}}" class="website-paper-collapse-item-text-input-section__input" />
+                    </div>
+                    <div>
+                        <label for="websiteCatchingRuleIsRegex">Regular expression format</label>
```
warpech: Can we change this label to one of the below:

In Atom it is called "Use Regex", in VS Code it is called "Use Regular Expression", in VS it is called "Use Regular Expressions".

PatrykSzwer:
I've managed to crash website with regex/uri like this:
`/[a-zA-Z]+/[a-zA-Z]+`

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/CatchingRulesPage.html
un-tone: Changed to "Use Regular Expression"

un-tone:
@PatrykSzwer, I guess you are talking about infinite loop when used nestable rule. I fixed it.

PatrykSzwer:
I'm not sure if It's expected behavior. Right now when I'm using regex that I pointed before, my result look like this:
image
I think It's not correct. My regex says, that URL for route has to be `/something/something` to be applied. 
`/[a-zA-Z]+/[a-zA-Z]`. But It's applied even when URI is: `/WebsiteEditor/surface/DyC/catchingrules` so `/something/something/something/something`. Except that, also this sidebar, created 2 times doesn't look good. 

It's somehow caused because of `+` sign. When I remove + everything works as expected.

un-tone:
It seems that your regex is incorrect and invalid. For `/something/something` route the correct regex should be `\/[a-zA-Z]+\/[a-zA-Z]+`

PatrykSzwer:
Hmm In that case maybe we should do something about incorrect regex? I'm not sure tho how could we do It, since probably typical checking with `try...catch` wouldn't throw an exception. Do You have any idea? Or should we just left it as it is?

un-tone:
I think we should left it as is. Otherwise we have to impose restrictions on regular expressions. I guess we don't want to do it.

TBH, I can't reproduce your case. @PatrykSzwer could you provide more detailed steps for it?

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -181,6 +173,19 @@ protected SurfacePage GetSurfacePage(WebTemplate template)
             return page;
         }
 
+        protected string GetWildCardUrl(string url)
+        {
+            var reg = new Regex(@"[/]\w*$", RegexOptions.IgnoreCase);
```
Mihaiii: Why only `\w`? Are we sure that we can't get a url that has, for example, "-"?

un-tone: Since we are replacing this feature in #191 and here, it doesn't matter anymore 😄

Mihaiii:
But I also agree that the user should be notified if he uses a wrong regex. I'm not aiming for a job done 100% right as we are talking about different engines, but you could validate it in JS and display a message like "The regex **might** be wrong". Maybe even be sure that it displays a slightly different message for (negative) lookbehind (as here might be the a lot of cases in which the notification is wrong) ?

I'm thinking that it might lead to real problems for clients if they are sure that they entered a valid regex and after a while it doesn't work as they expected and we all start digging around to see why it doesn't work. :)

PatrykSzwer:
@un-tone to reproduce my case, just go to sidebar/responsive surface and in catching rules make catching rule with catch url: `/[a-zA-Z]+/[a-zA-Z]`
image

un-tone:
@Mihaiii @PatrykSzwer such validation of regular expressions is a complex issue and I don't have a simple/fast solution for that. I guess currently nobody has. It means that should be implemented separately if we decide to do it.
Maybe we could create a test page for checking URLs and corresponding routes.

un-tone:
@PatrykSzwer I still can't reproduce the behaviour of your case. On my computer it falls the app down.

warpech:
We cannot detect incorrect regexpes but at least we can detect unparseable regexp, maybe let's do that?

un-tone:
@PatrykSzwer I fixed that bag `/[a-zA-Z]+/[a-zA-Z]`, please check it.

@warpech it sounds good, but regex validation with Regex doesn't suite such cases. So I didn't add such detecting when editing a route URI, also since the app doesn't have a proper user notification mechanism. But I added exception catching in WebsiteProvider middleware when regex creating and matching.

PatrykSzwer:
The bug is fixed, good job! 
I'm still not sure though if It's working correctly. I used this regex for test: `/[a-zA-Z]+e/[a-zA-Z]`
As I tested on `https://regexr.com/` this regex should catch URI for example `/people/persons` but shouldn't catch uri like `http://localhost:8080/WebsiteEditor/surface/Bs/catchingrules`. Unfortunately both are catched. But to make it more interesting,  regex like: `/[a-zA-Z]+e/[a-zA-Z]+s` is catched in `/people/persons` but not in any other not matching URI, except sidebarSurface page.

un-tone:
Why don't you escape `/` symbols? I think it should be `\/[a-zA-Z]+e\/[a-zA-Z]`

PatrykSzwer:
For some reason I though in our case we don't have to escape `/` symbol because We're parsing it and cuting URI. 
So one more thing now, when I escaped `/` symbols. Regex: `/[a-zA-Z]+e\/[a-zA-Z]+` should catch `/people/persons` but shouldn't `/WebsiteEditor/surface/Bs/catchingrules` (confirmed with `regexr.com`). Although It's still catched in both cases.

un-tone:
> I guess It can be catched by mistake because surface/Bs part is matched.

Not by mistake.
You should use `^(\/[a-zA-Z]+e\/[a-zA-Z]+)` expression for your case.

Issue #199, opened 2017-09-15
Exception during restarting WebsiteProvider when there is no surfaces.


PatrykSzwer:
### Steps to reproduce
1. Have WebsiteEditor, WebsiteProvider and SignIn running.
2. Login as an Admin user.
3. Go to [http://localhost:8080/WebsiteEditor/surfacegroups](http://localhost:8080/WebsiteEditor/surfacegroups)
4. Go to every surface and delete it.
5. Make sure that there is no surfaces left and restart only WebsiteProvider.
### Expected result
Everything should be running without exception.
### Actual result
Exception occurs:
![image](https://user-images.githubusercontent.com/25042256/30478267-b3eedc06-9a10-11e7-8840-614d268cfa7b.png)

##### Starcounter version  2.3.1.7180

cc @un-tone

PatrykSzwer:
Closing as done.

PR #201, opened 2017-09-15
Fixed way of deleting surface, to delete all childs along with it. #199

PatrykSzwer:
For #199


src/WebsiteEditor/ViewModels/GeneralPage.json.cs
```diff
@@ -64,6 +65,23 @@ void Handle(Input.Save action)
             this.Transaction.Commit();
         }
 
+        private void RemoveSurfaceElements()
+        {
+            foreach (var blendingPoint in this.BlendingPoints)
+            {
+                var pinningRulesOfBlendingPoint =
+                    Db.SQL<WebMap>("SELECT wm FROM Simplified.Ring6.WebMap wm WHERE wm.Section.Key=?",
```
un-tone: I guess you can use `WebSection.Maps` property instead of explicit query.

Issue #186, opened 2017-09-05
No refresh during switching between tabs of surface.


PatrykSzwer:
Issue created for bug mentioned here: https://github.com/StarcounterApps/Website/issues/176#issuecomment-327110473

### Steps to reproduce
1. Go to SidebarSurface page.
2. When General view is loaded click on Catching rules on navigation bar.
### Expected result
View is redirected to catching rules partial page.
### Actual result
URL is changed but nothing more is happening.
##### Starcounter version 2.3.1.7138

PatrykSzwer:
As part of investigating this, I've tried to replace links in navigation bar to triggers with puppet-redirect, but without results.

un-tone:
@PatrykSzwer thanks for the help. But I've already fixed the issue in `fix-nested-surface-render` branch 😄. Currently I'm working on tests for nested surfaces.

PatrykSzwer:
Ou sorry, but there is not any annotation to this branch so I though It's still unresolved. That's great!

un-tone:
Closing as fixed.

un-tone:
Reopening due to #202

un-tone:
Actually #205 and #206 fix the issue. Please review these PRs.

un-tone:
Fixed. Closing...

Issue #47, opened 2017-02-27
Add Workspaces to the Website app


warpech:
@malx122 represents use case in which the DOM state of blending point should be retained as it was done in Workspaces in the old Launcher.

A similar functionality should be implemented in Website.

Ideally it should allow caching multiple responses from the same app, as well as configuring the caching limit.

warpech:
I am aware that this requires more clarification before implementation. Assigning it to me for the time being.

warpech:
Discussed today between me and @malx122, there are at least two use cases for workspaces:

1. They make it possible to immediately switch to a view that was seen before
2. They retain state (such as position of scrollbars)

@malx122 uses the `AutoRefreshBoundProperties == false`, so that the server does not observe and push changes for data bound to the inactive workspaces.

tomalec:
What about Workspace micro app? That would provide a view, that could be blended into root,main area, any area, could blend in other apps responses, to keep them in the DOM instead of redirecting and track the scroll position.

un-tone:
AFAIK a work on this issue was started (and maybe finished) but I can't find a branch with that code changes. I wanted to check it and try with last other changes in WebsiteProvider app.

warpech:
I have pushed the proposed implementation. It needs careful review and testing with other apps (two tests are included).

I decided to rename the feature to "ViewKeeper". Hopefully that's more understandable?

un-tone:
Closing as done.
For the View Keeper settings I created a separate issue https://github.com/Starcounter/Website/issues/268/

PR #202, opened 2017-09-18
Revert "Fixed nested surface rendering"

un-tone:
Reverts StarcounterApps/Website#189

There is an error in [the line 179](https://github.com/StarcounterApps/Website/compare/develop...revert-189-fix-nested-surface-render?expand=1#diff-0b396d2f604aece8a3655c1601174c81L179) - I had to change store name there too.
But fixing it we can see that these changes doesn't fix the original issue #186.
I will return back the tests for this case when we find a new solution for the original issue.


warpech:
This error blocks implementation of workspaces (#47) so I agree it has to be prioritized. I think that the issue #186 has to be properly fixed before I submit my implementation of #47, otherwise, we are building a new feature on top of buggy behavior.

@un-tone if you think merging this PR to develop is a way to go then please merge. But we need to fix #186 with high priority anyway, so for me instead of merging a revert commit we can as well merge a working fix if we can find one.

un-tone:
>  instead of merging a revert commit we can as well merge a working fix if we can find one

Agree

un-tone:
Please look at the PR #205.

PR #205, opened 2017-09-18
Request handler mapping

un-tone:
Improved changes from #161.
Implemented what we discussed in https://github.com/Starcounter/Minutes/blob/23e6e8e38ee007d87017df83e7c9c148ba5b3627/BlendingTF/2017-09-14-blending_already_blended.md

Briefly: instead of trying to combine pinning and main contents (both can be already blended with other apps' responses) we map a requested URI from the WebsiteProvider's request handler to corresponding pinning content "caller" URI and unmap it from the response handler.


src/WebsiteProvider/Api/PinningHandlers.cs
```diff
@@ -79,6 +80,75 @@ public void Register()
...
+                {
+                    var urlToken = defaultSection.GetMappingToken(webUrl);
+                    if (Blender.IsMapped(requestUri, urlToken))
+                    {
+                        var blendInfo = Blender.ListByUri(requestUri).FirstOrDefault(x => x.Token == urlToken);
+                        if (blendInfo != null && blendInfo.IsFromUriConverterOn && !blendInfo.IsToUriConverterOn)
+                        {
+                            Blender.UnmapUri(requestUri, urlToken);
+                        }
+                    }
```
tomalec: WDYT about making 129-136 DRY with 139-146?

It's to unmap only those mapped as `Blender.MapUri(requestUri, _token_, true, false);` isn't it?

un-tone: Right. Fixed it. Thanks

PR #206, opened 2017-09-19
Changed back the session store name for SurfacePage

un-tone:
Changes from #205 fix the issue #186. In this PR I returned back the store name.

I will re-target this PR to `develop` after merging of the PR #205.


PatrykSzwer:
I think You can already merge #205 after change requested by Tomek, and go on with this one.

Issue #198, opened 2017-09-14
Implement real surface preview


warpech:
## Problem
 
It is hard to imagine how the surface looks like, specifically how are the blending points distributed visually.
 
## Proposed solution
 
For each surface, render a visualisation of it using an `<iframe>`. It might be scaled down to 10% or so. 
 
How to make this. The preview should have its own handler in WebsiteEditor that should inserts a view to each of the blending points. The view should just be an icon (🎯) and the name of the blending point.
 
For the "default" blending point, the text should also include suffix "(default)"
 
For example, the following preview view:
 
```html
<template>
<div class="surface-preview">🎯 $blendingPointName</div>
<style>
    .surface-preview{
            font-size: 24px;
            color: black;
    }
</style>
</template>
```
 
Rendered in a Holy Grail surface would look like:

![2017-09-14 14_37_53-content](https://user-images.githubusercontent.com/566463/30430379-5bba1010-995b-11e7-8917-e3230c6a9095.png)

Then, we can use such visualisation in:

- surface groups list (in a tooltip with a 200x200 px iframe)
- surface creator (though I am not sure here, the current diagrams look neat) 
- surface details (a new tab called "Preview")

@PatrykSzwer, @un-tone what do you think about this? I imagine this can be up to few days of work.

PatrykSzwer:
I think It's good to have feature. I actually though about It as a preview displayed in general tab on right side, not as a separated tab itself:
![prev](https://user-images.githubusercontent.com/25042256/30431281-634d7d82-995e-11e7-9015-5328bc9adad2.PNG)

 But It's just a easy to change positioning thing.

un-tone:
I've implemented a POC for this. Please look at https://github.com/StarcounterApps/Simplified/pull/78 and #208.

un-tone:
Closing as done

PR #208, opened 2017-09-20
Preview iframe for surfaces

un-tone:
For #198.


PatrykSzwer:
Looks neat! Just found out one thing.
When You create new, empty surface, and set it up same as e.g HolyGrailSurface (add same view URI and blending points) the preview is not showing anyway.

un-tone:
@PatrykSzwer good catch! Thanks! I fixed it in the last commit, please check.

PatrykSzwer:
Now It's working, good job! For me is actually good to go. 
I have only one concern. When You create new empty surface and just go into it, You will see in preview the default surface layer. I'm not if this wouldn't create some confusion in new users, who would thing e.g. that empty surface by default has topbar. But It's just my though, maybe we should ask @warpech, what do You thing about it?

un-tone:
I think I should do something with it, e.g. show an empty Preview or just hide a preview iframe

PatrykSzwer:
Ya, I'm thinking about maybe not empty but just some information, that there is nothing to show or something like this.

un-tone:
@PatrykSzwer I added message for cases when a surface is not configured properly.

PatrykSzwer:
Awesome! I would just change message to be something more precise, because "not configured properly" can mean anything. I would make it to something about invalid/empty View URI and/or no blending points to show. Something more about that.

un-tone:
@PatrykSzwer fixed

PatrykSzwer:
Now when I'm reading it: `Make sure that Surface view URI is correct and there are blending points created and refresh the page.` I think the `and refresh the page` part is redundant. WDYT? For me You can merge it after this last note.

Issue #209, opened 2017-09-22
Rename pinning rules to blending rules


PatrykSzwer:
After @tomalec suggestion We decided that in order to make pinning rules clearer, We should rename it to blending rules.

PatrykSzwer:
Closed as done.

PR #212, opened 2017-09-22
Rename pinning rules to blending rules

PatrykSzwer:
For #209 

Before I finish renaming I have one thing that I'm not sure about. In many cases word `pinning` is used but not in term of Pinning rules, but more as an action, as `pinning something`. I'm not sure if should I rename those to `blending something`, since It can create some confusion. 
Examples: 
pinningHandlers -https://github.com/StarcounterApps/Website/blob/a29f1bae7ba4f7897261128ad5c6e7caa92824b9/src/WebsiteProvider/Program.cs
PinPage, PinningPage in tests view models:
https://github.com/StarcounterApps/Website/tree/a29f1bae7ba4f7897261128ad5c6e7caa92824b9/test/WebsiteProvider_AcceptanceHelperTwo/ViewModels

Do You have an idea if Should I rename those or not?
cc: @warpech, @un-tone


un-tone:
I think you can rename it all, e.g. pinningHandlers to blendingHandlers, PinningPage to BlendingPage, PinPage to BlendingContentPage, etc

PatrykSzwer:
I've done last merge with develop. Please check if everything works as expected and review.

PR #213, opened 2017-09-23
Add asterisk in run.bat

warpech:
Explained in the commit message


Mihaiii:
I did this in https://github.com/StarcounterApps/Website/pull/120 , but indeed is not on `develop`. How come?

Issue #152, opened 2017-07-28
Change readme due to latest UI changes.


PatrykSzwer:
Readme should be changed to be accurate with current UI. Tabs should be re-described.

warpech:
@PatrykSzwer do you think this can be closed as done now?

PatrykSzwer:
Yes, closing as done.

PR #221, opened 2017-09-28
Enhance README

warpech:
Implements #152. More info in the commit message


README.md
```diff
@@ -10,58 +12,39 @@ The app requires a user to be signed in in order to work. This can be done with
...
+A route is a condition by which the surface is displayed. Routes are matched against the HTTP requests from the web browser. If the request matches a route, the response view will be decorated by the surface. Routes can be very specific, for example only display the surface if the request URI starts with `/people` and the user agent is iPad Safari. Routes can also be general when you want the surface to be displayed for all requests. 
 
 Routes support Starcounter routing format of URLs. So, for example, you can use the following route URIs:
 
 * `/application/resource/{?}`
 * `/application/resource/{?}/edit`
 * `/application/query?{?}`
 
-A route can be **Final**, meaning that not any other routes will be applied additionally. If a route have an empty **Route URI** value, it will be forced (only first one) on any JSON response from any Starcounter app. In case if not any route was matched for a response, it will be loaded "as-is".
+A route can be **Final**, meaning that no any other routes will be applied additionally. If a route has an empty **Route URI** value, it will be forced (only first one) on every JSON response from any Starcounter app.
```
PatrykSzwer: Since we get rid of word "Final" in UI, shouldn't we also remove it from documentary? And replace it with Nestable/Allow route nesting?

README.md
```diff
@@ -1,6 +1,8 @@
 # WebsiteEditor and WebsiteProvider
 
-Suite of two apps allowing to create surfaces with blending points and attach the result from different applications in the blending points.
+WebsiteEditor is a graphical tool for Starcounter system owners that allows managing blending rules for the user interface. You can benefit from it even if your system consists of only one app, but the true benefit shows when you have multiple apps.
```
PatrykSzwer: I don't like word graphical here. Maybe we could replace it with something different? Like configuration tool, or managing tool. Also this first sentence sounds like websiteEditor's purpose is only to manage blending rules, but there are also routes, blending points, surfaces etc.

warpech: I don't want the sentence to sound overwhelming... Maybe:

> WebsiteEditor is a tool for Starcounter system owners to manage blending of the user interface.

README.md
warpech: Actually I just pasted this part from the current README, sorry for not paying attention. I will rephrase that.

README.md
PatrykSzwer: Exactly what I was thinking about!

Issue #222, opened 2017-09-29
Default surface from template is created without blending points


warpech:
### Steps to reproduce

1. Start WebsiteEditor (branch `develop`), SignIn (branch `develop`)
2. Call `http://localhost:8080/website/resetdata` to load default mappings
3. Go to `http://localhost:8080/signin/signinuser` to generate the admin user and sign in
4. Go to `http://localhost:8080/websiteeditor/surfacegroups`
5. Create a new surface from the "Default" template
6. Click on "New default template" -> "Blending points"

### Expected result

The new surface should have two blending points with the bottom one marked as `default`. 

### Actual result

The new surface has no blending points.

##### Starcounter version 2.3.1.7294

PatrykSzwer:
Closing as done.

PR #225, opened 2017-09-29
Added default blending points to surface created from default template

PatrykSzwer:
For #222


src/WebsiteEditor/Helpers/DataHelper.cs
```diff
@@ -96,7 +96,7 @@ public void GenerateDefaultSurface()
             var topbar = new WebSection
             {
                 Template = surface,
-                Name = "TopBar",
+                Name = "Topbar",
```
warpech: In README.md it is called "TopBar". Please sync that.

warpech:
The tests are failing, can you check that?

Issue #227, opened 2017-09-29
SurfaceGroupsPage is not DRY


warpech:
The code in `SurfaceGroupsPage.html`, `json` and `json.cs` is duplicated for grouped and ungrouped surfaces in lines:

- https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html#L9-L56
- https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html#L60-L108

Ungrouped surfaces should be seen as belonging to a group that is null. There is no need for the code to be duplicated.

un-tone:
Closing as done.

PR #228, opened 2017-10-02
Made SurfaceGroupsPage DRY

PatrykSzwer:
For #227


warpech:
@un-tone can you pls review?

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/SurfaceGroupsPage.html
```diff
@@ -7,13 +7,20 @@
         <div class="website-groups">
             <template is="dom-repeat" items="{{model.SurfaceGroups$}}">
                 <h4>
-                    <input type="text" value="{{item.Name$::change}}" class="website-surface-group-name"/>
+                    <template is="dom-if" if="{{!isUnassignedGroup(item.Name$)}}">
+                        <input type="text" value="{{item.Name$::change}}" class="website-surface-group-name" />
+                    </template>
+                    <template is="dom-if" if="{{isUnassignedGroup(item.Name$)}}">
```
un-tone: Let's use another field, not the `Name`. Maybe it's better to create a separate flag for it, e.g. `IsUnassigned`. Otherwise, user can set name "Unassigned" for a group and will not able to change it after.

un-tone: If you add such field, the `isUnassignedGroup` function can be removed also.

PatrykSzwer: Ok, good idea. Will do.

Issue #200, opened 2017-09-15
After deleting ResponsiveSurface or SidebarSurface redirecting doesn't work


PatrykSzwer:
### Steps to reproduce
1. Have WebsiteEditor, WebsiteProvider and SignIn running.
2. Login as an Admin user at http://localhost:8080/signin/signinuser
3. Go to http://localhost:8080/WebsiteEditor/resetdata
4. Go to http://localhost:8080/WebsiteEditor/surfacegroups
5. Go to ResponsiveSurface or SidebarSurface and remove it.
### Expected result
User is redirected to `/surfacegroups` page, as It's happening during deleting e.g HolyGrailSurface
### Actual result
Redirect doesn't work and user stays on empty page.

##### Starcounter version 2.3.1.7180

cc @un-tone 

BTW, 200th issue in website, Let's celebrate! 🎆 🎉

un-tone:
I think this issue is related to #186

warpech:
Bonus problem: After doing steps from the OT, do:

- Go to http://localhost:8080/WebsiteEditor/routing

## Expected

You should see a list of routes defined in WebsiteEditor

## Actual

You get `System.NullReferenceException: Object reference not set to an instance of an object`  and **the whole database crashes**. For me, the fact that WebsiteProvider can crash the entire database seems like a problem in Starcoutner 2.3 that @chrhol and @miyconst might want to take a look at.

Error log shows two errors:

```
System.Exception: ScErrUnhandledException (SCERR1069): An unhandled exception was detected.
Version: 2.3.1.7294.
Help page: https://docs.starcounter.io/v/2.3.1/?q=SCERR1069.
   at StarcounterInternal.Hosting.Processor.<>c__DisplayClass3_0.<RunTaskNoScheduling>b__0() in C:\TeamCity\BuildAgent\work\sc-8734\Level1\src\Hosting\Starcounter.Hosting\Processor.cs:line 138
   at Starcounter.Internal.StarcounterEnvironment.RunWithinApplicationInternal(String desiredAppName, Action action) in C:\TeamCity\BuildAgent\work\sc-8734\Level1\src\Starcounter.Internal\Mock\StarcounterEnvironment.cs:line 142
   at StarcounterInternal.Hosting.Processor.RunTaskNoScheduling(Action userAction, String appName) in C:\TeamCity\BuildAgent\work\sc-8734\Level1\src\Hosting\Starcounter.Hosting\Processor.cs:line 121
[...]

PR #229, opened 2017-10-03
Delete all surface stuff when deleting it

un-tone:
For https://github.com/StarcounterApps/Website/issues/200#issuecomment-332985614


src/WebsiteEditor/ViewModels/GeneralPage.json.cs
```diff
@@ -77,9 +77,20 @@ private void RemoveSurfaceElements()
...
             }
+            foreach (WebUrl route in this.Surface.Data.Pages)
+            {
+                foreach (var webProperty in route.Properties)
+                {
+                    webProperty.Delete();
+                }
+                route.Delete();
+            }
+            foreach (var section in this.Surface.Data.Sections)
```
PatrykSzwer: Why are You deleting Sections (blending points) here? Above this there is already removing for them and I don't see that after deleting surface there are some still remains.

un-tone: Thanks! You are right. Fixed it.

Issue #216, opened 2017-09-26
Get rid of Key and ObjectID properties using


un-tone:
In according to this issue https://github.com/Starcounter/Home/issues/252 and decisions from https://github.com/Starcounter/level1/issues/4210 we should replace all occurrences of using `Key` and `ObjectID` properties with `ObjectNo`.

warpech:
Even better, replace it with `DbHelper.FromID(DbHelper.Base64DecodeObjectID(id))` and pass objects in queries instead of ids.

un-tone:
Closing as done.

PR #230, opened 2017-10-03
Replace key and objectID usage

PatrykSzwer:
for #216 

During replacing queries with DbHelper I found out one problem. There are places where key value for query is expected to be null in some cases. E.g. here https://github.com/StarcounterApps/Website/blob/develop/src/WebsiteEditor/ViewModels/GeneralPage.json.cs#L34 during setting default blending point, user can choose None, and this is read as an empty key. 
Solution with query has no problem with it, but DbHelper.Base64DecodeObjectID() is throwing errors when argument in Its methods is null/empty string.
I know that the easiest way would be to add ifs and checks for nullOrEmpty string, but that generates code and I don't know is it best solution. Maybe you will have some better ideas how to overpass this.

cc @warpech, @un-tone


src/WebsiteEditor/ViewModels/GeneralPage.json.cs
```diff
@@ -19,7 +19,7 @@ public void RefreshData()
                 throw new InvalidOperationException("Surface key is empty.");
             }
 
-            this.Surface.Data = Db.SQL<WebTemplate>("SELECT t FROM Simplified.Ring6.WebTemplate t WHERE t.Key = ? ORDER BY t.Name", SurfaceKey).First;
+            this.Surface.Data = DbHelper.FromID(DbHelper.Base64DecodeObjectID(SurfaceKey)) as WebTemplate;
```
warpech: This is not equivalent. Unfortunately, here I think that here you have to use `ObjectNo`

src/WebsiteEditor/ViewModels/GeneralPage.json.cs
```diff
@@ -33,8 +33,7 @@ public void RefreshData()
 
         void Handle(Input.DefaultBlendingPointKey action)
         {
-            var currentDefaultBlendingPoint =
-                Db.SQL<WebSection>("SELECT s FROM Simplified.Ring6.WebSection s WHERE s.Key = ?", action.OldValue).First;
+            var currentDefaultBlendingPoint = Db.SQL<WebSection>("SELECT s FROM Simplified.Ring6.WebSection s WHERE s.Key = ?", action.OldValue).First;
```
warpech: Here nothing has changed. I think that you have to use `ObjectNo`

src/WebsiteEditor/ViewModels/RoutingPage.json.cs
```diff
@@ -60,9 +60,7 @@ void Handle(Input.CancelChangesTrigger action)
 
         void Handle(Input.CreateModalTrigger action)
         {
-            var surface = Db
-                .SQL<WebTemplate>("SELECT wt FROM Simplified.Ring6.WebTemplate wt WHERE wt.Key = ? ORDER BY wt.Name",
-                    this.ChoosenSurfaceKey).FirstOrDefault();
+            var surface = DbHelper.FromID(DbHelper.Base64DecodeObjectID(ChoosenSurfaceKey)) as WebTemplate;
```
warpech: This is not equivalent. Unfortunately, here I think that here you have to use `ObjectNo`

src/WebsiteEditor/ViewModels/BlendingPointsPage.json.cs
```diff
@@ -21,7 +21,7 @@ public void RefreshData()
             }
 
             this.BlendingPoints.Clear();
-            this.Surface.Data = Db.SQL<WebTemplate>("SELECT t FROM Simplified.Ring6.WebTemplate t WHERE t.Key = ? ORDER BY t.Name", SurfaceKey).First;
+            this.Surface.Data = DbHelper.FromID(DbHelper.Base64DecodeObjectID(SurfaceKey)) as WebTemplate;
```
warpech: This is not equivalent. Unfortunately, here I think that here you have to use `ObjectNo`

src/WebsiteEditor/ViewModels/GeneralPage.json.cs
PatrykSzwer: That was a bug, this order by shouldn't be there.

src/WebsiteEditor/ViewModels/RoutingPage.json.cs
PatrykSzwer: Same as above.

src/WebsiteEditor/ViewModels/BlendingPointsPage.json.cs
PatrykSzwer: Same as above.

PatrykSzwer:
> Otherwise you need to keep the previous queries, just change Key to ObjectNo

To use ObjectNo I need to get it, by DbHelper.Base64DecodeObjectID(). To use DbHelper.Base64DecodeObjectID() argument cannot be empty. There are cases where it is empty. That's the problem.

warpech:
Then I can see not better option as to put it in a conditional clause. Unless @un-tone has some better idea. But I guess it will be better to look at the code once you've done it.

warpech:
@PatrykSzwer started his vacation today.

@un-tone should we merge it?

un-tone:
I think yes if you approve it

PR #197, opened 2017-09-14
194 routing at home page

PatrykSzwer:
For #194 .

This branch was based on branch from PR #196 and has to be proceed after #196 is merged.

PR #236, opened 2017-10-11
Restore "Use Regular Expression"

warpech:
Problem: When PR #197 and #190 were merged, the checkbox Use Regular Expression was lost during merge

Solution: Restore the checkbox Use Regular Expression


PatrykSzwer:
Good catch! I also committed my proposition of better arrangement of route options. Please take a look and if You find It more readable same as me, please merge.

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/RoutingPage.html
```diff
@@ -60,8 +60,11 @@
                     </div>
                 </div>
                 <div>
-                    <label for="websiteRouteDisabled">Disabled</label>
-                    <input type="checkbox" id="websiteRouteDisabled" checked="{{route.Disabled$::change}}"/>
+                    <label for="websiteCatchingRuleIsRegex">Use Regular Expression</label>
```
warpech: Can you pls change "regular expression" to be lowercase to be consistent with other labels?

PatrykSzwer: Done

warpech:
Lgtm

Issue #239, opened 2017-10-16
Visual improvements for WebsiteEditor


PatrykSzwer:
After long discussion with @warpech, We decided to make few small changes to improve user experience for Website Editor app:

1. Rename unassigned surfaces to ungrouped surfaces.
2. Make iframe for preview wider so that holy grail looks correct, and make Preview header as a link to new window.
3. Rename Route URI in blending rules to route.
4. Fix Route URI (in Blending rules) content - add headers from routes view.
5. Replace green + button for adding surface with more intuitive "add surface" button placed on the end of surfaces.
6. Merge "URI" and "Headers" in route header.
7. Rename Route nesting enabled to Surface nesting in route header.
8. First name in Route URI in blending point should be "All Routes"

warpech:
Closing as done in #241

PR #241, opened 2017-10-17
239 visual improvements

PatrykSzwer:
For #239


un-tone:
image

Blending rule row looks roughly. Can you align the elements?

-----

Also headers part of Route element looks better if wrap it in braces, WDYT?
image

src/WebsiteEditor/Helpers/AppHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Simplified.Ring6;
+
+namespace WebsiteEditor.Helpers
+{
+    public static class AppHelper
+    {
+        public static string GetHeadersString(IEnumerable<WebHttpHeader> headers)
```
un-tone: What about to make this method as extension method?

PatrykSzwer: Love the idea! Consider it done!

Issue #240, opened 2017-10-16
Routes ordering


PatrykSzwer:
Right now It's impossible to predict how multiple nested surfaces will behave. We need to implement ordering for routes and make surfaces nesting conditional to it. 

This issue is related to https://github.com/StarcounterApps/Simplified/pull/77 and https://github.com/Starcounter/Website/issues/195.

un-tone:
Closing as done.

PR #243, opened 2017-10-18
Routes ordering

PatrykSzwer:
For #240 

I've implemented controlling and all visual aspects of routes ordering. Now It only lacks implementation in WebsiteProvider. @un-tone please make your changes/commits here, to this PR.


PatrykSzwer:
As we discussed in a meeting with @un-tone we have to consider one thing here.
Right now, website provider is looking for matching routes in this order:
1. Specific route, like /something/someone/key
2. Wildcart route, like /something/someone/{?}
3. Regex route, like /something/someone/[a-zA-Z]+e
4. and finally route-all route.

We have to decide if route ordering should be above this order, or it should first follow that and then route ordering set manually. Second solution would be much easier to implement, but I think user would expect first scenario to happen. @warpech WDYT?

warpech:
I am telling everyone that WebsiteProvider applies the rules in the order in which they are displayed, from top to bottom. I know this is not the case yet, but I think it is a reasonable goal.

If you agree, then the automatic ordering listed in https://github.com/Starcounter/Website/pull/243#issuecomment-338637499 should be abandoned.

un-tone:
The changes are ready for WebsiteProvider's part.

warpech:
Thanks! @PatrykSzwer can you review pls?

@un-tone did you review the WebsiteEditor part?

un-tone:
Yes. It's okey, but requires some visual changes, maybe drag&drop for ordering routes.

PatrykSzwer:
Drag&drop feature has separated issue. Lets make all related visual improvements there, and here left only functionality itself.

PatrykSzwer:
I've tested this and I have couple of thoughts.

1. I don't think that ordering is working correctly right now. I've set two surfaces/routes that should work for same uri, `/people/persons`, tobbarSurface, our default one, and holyGrailSurface. When ordering is set to same number for both of them, only tobbarSurface is showing up. When I've set tobbarSurface for 1 and holyGrail for 0, I can see that both of them are working. When I've set TobbarSurface on 0 and HolyGrail for 1, again only Tobbar is visible. I'm not sure but I think that in situation where nesting is allowed, and only 2 surfaces with routes are matching the URI, We should show both of them nested. Order could decide which one should be nested in which (I'm not sure on that one, @warpech WDYT?).

2. Second thing, don't know what about You but I think that ordering from 0 is confusing, I think we should change it to start ordering from 1. I'll also add a sentence to Routing description after We agree and finish this functionality in 100%.

un-tone:
@PatrykSzwer why have you changed this line https://github.com/Starcounter/Website/pull/243/commits/c2b0b677c1ba28bdbc0070a4824a98a04198fbae#diff-791d27d90e41e68a4dbfe0c5512476a6R115 ? It is a route-all, should have a low priority.

PatrykSzwer:
Though It was just a mistake/test value that wasn't removed. I'll revert that, sorry.

un-tone:
Some bugs:
1. After deleting all surfaces I'm getting an empty page (with `http://localhost:8080/WebsiteEditor/settings` URL). Works only if any surfaces existed. In the code it set redirection to `/WebsiteEditor/surfacegroups`, I think we should stay on `/WebsiteEditor/settings` page.
1. Importing doesn't work for me. I've successfully exported a dump and downloaded it. When importing this dump nothing is happening. I haven't gotten any user notification about importing status.

un-tone:
I agree with your comment https://github.com/Starcounter/Website/pull/243#issuecomment-341474414 and will fix

un-tone:
Fixed. Please review.

PatrykSzwer:
There are some new merge conflicts, please take a look.

un-tone:
Fixed the conflicts.

PatrykSzwer:
Looks like everything's working perfectly! I'll add description line about routing and after reviewing it You can merge those changes.

un-tone:
Looks good. Merging...

PR #245, opened 2017-10-19
Tests improvement

PatrykSzwer:
This PR is for keeping consistence between kitchensink and website tests. It's related to https://github.com/Starcounter/KitchenSink/issues/190 and https://github.com/Starcounter/KitchenSink/pull/191


un-tone:
One question - is it okey that we run tests in maximized window only for Chrome, but not for others?

PatrykSzwer:
You overrun me, I was just changing this. I was waiting for https://github.com/Starcounter/KitchenSink/pull/191 to be done before I go with this PR further.

Issue #244, opened 2017-10-19
Use Db.FromId


warpech:
In PR #230 an API `DbHelper.FromID` was used, but since 
https://github.com/Starcounter/level1/pull/4218 in Starcounter there is much simpler API: `Db.FromId`. Please use that one one instead.

https://docs.starcounter.io/guides/database/object-identity-and-object-references#retrieving-from-object-identity

PatrykSzwer:
Closing as done.

PR #246, opened 2017-10-19
Use Db.FromId

PatrykSzwer:
For #244


src/WebsiteEditor/Api/Authorization/MasterPageMiddleware.cs
```diff
@@ -28,7 +28,7 @@ public Response Run(RoutingInfo routingInfo, Func<Response> next)
             if (!isSurfaceGroupsPage && routingInfo.Arguments.Any())
             {
                 var surfaceKey = routingInfo.Arguments?[0];
-                master.Surface.Data = DbHelper.FromID(DbHelper.Base64DecodeObjectID(surfaceKey)) as WebTemplate;
+                master.Surface.Data = Db.FromId<WebTemplate>(DbHelper.Base64DecodeObjectID(surfaceKey));
```
Mackiovello: You don't need `Base64DecodeObjectID`, `Db.FromId<WebTemplate>(surfaceKey)` is enough.

PatrykSzwer: Thats even cooler! Consider it Done!

Issue #247, opened 2017-10-19
Remove the support for executing blending rules for the default blending point


warpech:
For the reasons explained in:

- https://github.com/Starcounter/Website/issues/237 (two tests fail after `Blender.MapUri` improvements)
- https://github.com/Starcounter/Website/issues/238 (WebsiteProvider uses costly calls to `Blender.MapUri` and `Blender.UnmapUri` at runtime in middleware)

We decided to do the following:

1. remove the feature of blending additional content to the "default blending point"
    - remove the code in WebsiteProvider that does that
    - change the current tests in WebsiteProvider to check that it does NOT work
2. whenever a user tries to set a blending rule to the default blending point in WebsiteEditor, we will show a warning that blending rules cannot be added to the default blending points.
    - I imagine this can be a warning sign with the description in a tooltip
    - a user can still define such blending rule, it will just not work

@PatrykSzwer please implement the changes in WebsiteEditor and WebsiteProvider. @un-tone will review both.

PatrykSzwer:
Done, closing.

PR #248, opened 2017-10-20
Removed support of blending rules for the default blending point

PatrykSzwer:
For #247 

I finished first point listed in the issue description. I'll now implement second point, in a meantime please check if my changes in websiteProvider are good to go.

EDIT: Seems like I've naively changed only previously failing tests, and I need to change all of them. I'm on it now.


PatrykSzwer:
I've changed rest of the tests due to this change. @un-tone please check if those changes are ok for you.

PatrykSzwer:
Yes of course. I've also added requested warning for user, which says that blending rules for default blending point won't work. I was trying to do it with tooltip but every try ended with something more ugly than just simple warning text next to the button.

Please do final check and I'll merge this.

un-tone:
I think it will look better if you shift the warning a little to the right
image

un-tone:
Another thing with blending of such rules in the default points - we still register them in the `WebsiteProvider.Api.BlendingHandlers` though them are not used really. There are 2 ways I see:

1. We leave the registering since it doesn't cost much, I guess. Also it doesn't require to make changes in the code. But on the other side I think it is a bad practice and it is not obvious for other developers why we map unused rules.
2. Do not map rules for default points. Here we will have to handle changes of surface's default points and do unmapping of previous default point's rules and do mapping of new ones. It will complicate more the already over-complicated algorithm of `BlendingHandlers`.

I think we should go to the 1st way currently, but also create a new issue for the 2nd way for the future. After discussing with @PatrykSzwer I see that we should simplify some parts of `BlendingHandlers`, and after that we could implement the 2nd way.

@PatrykSzwer @warpech WDYT?

PatrykSzwer:
I agree with You, let's not bother with this now, create issue for it and improve it during simplifying Website Provider process.

PatrykSzwer:
> I think it will look better if you shift the warning a little to the right

I've shifted it, but just a little, I want it to be obvious that is related to that button.

warpech:
> I think we should go to the 1st way currently, but also create a new issue for the 2nd way for the future. After discussing with @PatrykSzwer I see that we should simplify some parts of BlendingHandlers, and after that we could implement the 2nd way.

I agree!

PatrykSzwer:
I've added default label next to a default blending point name. Please take a look.

src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels/BlendingPointsPage.html
```diff
@@ -23,7 +23,12 @@
         <template is="dom-repeat" items="{{model.BlendingPoints}}" as="blendingPoint">
             <paper-collapse-item>
                 <div slot="header" class="website-paper-collapse-item-header">
-                    <div class="website-paper-collapse-item-header__label">{{blendingPoint.Name$}}</div>
+                    <div class="website-paper-collapse-item-header__label">
+                        {{blendingPoint.Name$}}
+                        <template is="dom-if" if="{{blendingPoint.Default$}}">
+                            <b>(Default)</b>
```
warpech: Typically, using `<strong>` is advised in HTML5. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

PatrykSzwer: Done.

PatrykSzwer:
@warpech please do final check and if everything is ok, I'll merge those changes.

Issue #217, opened 2017-09-26
Remove ignore tags above disabling stuff tests


PatrykSzwer:
After https://github.com/Starcounter/Home/issues/251 is fixed, Ignore tags above tests should be removed and tests should be passing.

un-tone:
Closing as done.

PR #253, opened 2017-10-26
Remove ignore tags from DisablingStuff tests

PatrykSzwer:
#217 

Since Starcounter/Home#251 is done, and tests passes, We can remove ignore tags already.


PatrykSzwer:
I've merged develop with #248 changes included to this branch, and removed last tag. Please check if everything is working correctly.

un-tone:
There is some issue with building in TC (for example http://teamcity/viewLog.html?tab=buildLog&buildTypeId=Apps_Website_Develop&buildId=35331 works more than 25h). I will reboot the agent machine and check results.

un-tone:
FYI, I created the issue https://github.com/Starcounter/AdminTrack/issues/363 for the TeamCity problem.

PR #256, opened 2017-10-26
Fixed surface preview

un-tone:
For https://github.com/Starcounter/Website/issues/250.


PatrykSzwer:
Fix looks ok, just would be awesome if You could provide few words about it, what and how It fixed. Also if It's possible, You can try to fix that: when you change default blending point and click save, preview information about default blending point is not changed. But It's nice to have thing, since all you have to do now, is refresh page, so It's not necessary.

un-tone:
> few words about it, what and how It fixed

The fixes relate to changes in handling of specifics URIs by Blender: https://docs.starcounter.io/guides/blending/blending#blending-specific-uris

> when you change default blending point and click save, preview information about default blending point is not changed

Sure, I'll try to fix it

un-tone:
Implemented iframe refreshing.

src/WebsiteEditor/ViewModels/GeneralPage.json.cs
```diff
@@ -73,6 +73,11 @@ void Handle(Input.DeleteTrigger action)
         void Handle(Input.Save action)
         {
             this.Transaction.Commit();
+            
+            // hack for triggering iframe refreshing
+            this.PreviewUrl = this.PreviewUrl.StartsWith("/websiteeditor/")
+                ? $"/WebsiteEditor/surface/{SurfaceKey}/preview"
+                : $"/websiteeditor/surface/{SurfaceKey}/preview";
```
PatrykSzwer: I don't get the sense here, why even checking anything, since anyway you're setting previewUrl on ` $"/WebsiteEditor/surface/{SurfaceKey}/preview`.? It's working same way as would work:
` this.PreviewUrl = $"/WebsiteEditor/surface/{SurfaceKey}/preview";`

un-tone: > It's working same way as would work:
this.PreviewUrl = $"/WebsiteEditor/surface/{SurfaceKey}/preview";

Only first time, right?

PatrykSzwer: Oooou I didn't saw that difference in upper-case letters on first time. Sorry about that, now I get that hack, and why It's working.

Issue #252, opened 2017-10-25
Fix authorization library using


un-tone:
Current implementation doesn't let simply add another page types - it works well only for ones implementing `IKnowSurfacePage`.

Also it is incorrect re-creating of `CurrentPage` here https://github.com/Starcounter/Website/blob/16beaa01bbbf241dceb1a23c48c37728d0774224/src/WebsiteEditor/Api/Authorization/MasterPageMiddleware.cs#L42 since it already assigned in the code a little above https://github.com/Starcounter/Website/blob/16beaa01bbbf241dceb1a23c48c37728d0774224/src/WebsiteEditor/Api/Authorization/MasterPageMiddleware.cs#L25

Moreover this code looks redundant since we use authorization library components https://github.com/Starcounter/Website/blob/33e85a23e3dd443d42f310d766469bc36a1b3ff6/src/WebsiteEditor/Api/MainHandlers.cs#L80-L86

warpech:
Is this something that @PatrykSzwer can fix in WebsiteEditor or is it a fault of the Authorization library?

PatrykSzwer:
I can try to partly fix it, but I can't promise that I can handle all of this problems.

PatrykSzwer:
Closing as done.

PR #263, opened 2017-11-06
Reimplementation of authorization library

PatrykSzwer:
For #252


PatrykSzwer:
Conflicts fixed.

src/WebsiteEditor/ViewModels/MasterPage.json.cs
```diff
@@ -10,34 +10,12 @@ partial class MasterPage : Json
     {
         protected string AllowedSystemUserGroup = "Admin (System Users)";
 
-        public void RefreshCurrentPage(string partialPageUrl)
-        {
-            this.PartialUrl = partialPageUrl;
-            this.RefreshCurrentPage();
-        }
-
         public void RefreshCurrentPage()
```
un-tone: We should do refreshing page here or just remove the method

src/WebsiteEditor/ViewModels/MasterPage.json.cs
```diff
@@ -10,34 +10,12 @@ partial class MasterPage : Json
     {
         protected string AllowedSystemUserGroup = "Admin (System Users)";
```
un-tone: This field is not used

src/WebsiteEditor/Api/Authorization/AuthEnforcementProvider.cs
```diff
@@ -41,7 +41,7 @@ private static AuthorizationEnforcement InitializeAuthEnforcement()
         private static AuthorizationRules GetAuthorizationRules()
         {
             var rules = new AuthorizationRules();
-            rules.AddRule(new ClaimRule<ShowSurfaceGroups, SystemUserClaim>((claim, permission) => claim.SystemUser.Name == "admin"));
+            rules.AddRule(new ClaimRule<ShowSurfaces, SystemUserClaim>((claim, permission) => claim.SystemUser.Name == "admin"));
```
un-tone: Maybe instead of checking if the user name is "admin", we should check if the SystemUser is assigned to an admin group like it have been done in Calendar app, WDYT?

un-tone:
Additional notes:

1. Logging in/out does not trigger refreshing of current page.
2. View-models for all WebsiteEditor's pages are not namespaced. So I'm getting the following warning in browser console:
```
Your view-model is probably not namespaced!
Proxy {Html: "/WebsiteEditor/viewmodels/SurfaceGroupsPage.html", SurfaceGroups$: Proxy, CreateGroupTrigger$: 0, ChoosenTemplate$: 0, ConfirmTrigger$: 0} 
We no longer support those.
Please make sure it's a result of `Self.GET` on serverside. If it's not a result of blending, then replace `<starcounter-include view-model="{{viewModel}}">` with  `<div><template is="imported-template" model="{{viewModel}}" href="{{viewModel.Html}}"></template></div>`.
```
I think we should consider this requirement while we are re-implementing this part. But I'm not sure if it is enough for the app to replace `starcounter-include` with `template is="imported-template"`. Otherwise we should define partial URIs like it have been done in Images app.

src/WebsiteEditor/ViewModels/MasterPage.json.cs
PatrykSzwer: I'll remove this part here, and create a new issue after this is merged, for adding refreshing page for sign in/out with new authorization implementation. WDYT?

PatrykSzwer:
> Logging in/out does not trigger refreshing of current page.

See comment above for this part.

>View-models for all WebsiteEditor's pages are not namespaced.

I think we will end with Partial URIs solution. AFAIR replacing `starcounter-include` with `imported-template` can cause additional troubles, like not working `declarative-shadow-dom`.

PatrykSzwer:
I've done implementing Partial URIs solution. Please do final check.

src/WebsiteEditor/Api/CommitHooks.cs
```diff
@@ -33,9 +33,6 @@ protected void RefreshSignInState()
                     return;
                 }
 
-                var page = session.Store[nameof(MasterPage)] as MasterPage;
-
-                page?.Transaction.Scope(() => page.RefreshCurrentPage());
```
un-tone: This method `RefreshSignInState` should not be called since it does nothing useful.

PatrykSzwer: Should I remove the entire CommitHooks file? Since there is nothing more than calling for that method, and we will probably need that back when We're going to implement refreshing back.

un-tone: It's up to you. You can remove the file or just comment out the code if you will need it soon.

PatrykSzwer: I've commented it. Take a final look.

Issue #255, opened 2017-10-26
Generate the DOM part of the surface views automatically


warpech:
One of the remaining weak points of WebsiteEditor & WebsiteProvider is that your surface needs a static HTML file that represents a view of that surface. For example: https://github.com/Starcounter/Website/blob/6c5d6ececa62b28a8a242cec14351893215970a3/src/WebsiteProvider/wwwroot/WebsiteProvider/surfaces/TopbarSurface.html

This file should be created on the fly by a handler that returns something like this:

```html
<link rel="import" href="/sys/polymer/polymer.html">
<template>
    <template is="dom-bind">
        <starcounter-include slot="websiteeditor/topbar" view-model="{{model.Sections.Topbar}}" class="website-topbarsurface-topbar"></starcounter-include>
        <starcounter-include slot="websiteeditor/main" view-model="{{model.Sections.Main}}"></starcounter-include>
    </template>
    <template is="declarative-shadow-dom">
        <slot name="websiteeditor/topbar"></slot>
        <slot name="websiteeditor/main"></slot>
    </template>
</template>
```

Everything that is a difference between the above snippet and the current content of https://github.com/Starcounter/Website/blob/6c5d6ececa62b28a8a242cec14351893215970a3/src/WebsiteProvider/wwwroot/WebsiteProvider/surfaces/TopbarSurface.html should be a **custom composition** that is stored in the database like a custom composition of any other view of a Starcoutner app.

@PatrykSzwer can you try to prototype this? We can discuss this in person in more detail.

PatrykSzwer:
Closing as done. 🎆 🎉

PR #270, opened 2017-11-09
Generated DOM of the surfaces

PatrykSzwer:
For #255 

Be caution, that after this change, compositionProvider will be necessary for website to work properly.


PatrykSzwer:
I found out couple of small issues and that tests are failing. I'm currently working on it.

PatrykSzwer:
All issues should be fixed.

src/WebsiteEditor/Helpers/DataHelper.cs
```diff
@@ -119,16 +121,83 @@ public void GenerateTopbarSurface()
             new WebMap { Section = topbar, ForeignUrl = "/signin/user", SortNumber = 1 };
 
             GenerateSurfacePreview(surface, new[] { topbar, main });
+
+            new HTMLComposition
+            {
+                Key = $"/sc/htmlmerger?WebsiteProvider={surface.Html}",
+                Value = @"<style>:host{display:block}</style>
```
un-tone: Let's store the compositions content as embedded resource, like it has been done in https://github.com/Starcounter/inside.starcounter.com

src/WebsiteEditor/Helpers/DataHelper.cs
```diff
@@ -16,6 +16,7 @@ public void ClearData()
                 Db.SlowSQL($"DELETE FROM {typeof(WebSection).FullName}");
                 Db.SlowSQL($"DELETE FROM {typeof(WebTemplate).FullName}");
                 Db.SlowSQL($"DELETE FROM {typeof(WebTemplateGroup).FullName}");
+                Db.SlowSQL($"DELETE FROM {typeof(HTMLComposition).FullName}");
```
un-tone: Should we clear compositions in this way? What if we have ready compositions for People+Images+UserAdmin, etc? I don't think I want to lose all my compositions just calling `/webisteeditor/reset`. I think we should delete only compositions for surfaces, e.g. by composition keys.

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -55,6 +57,25 @@ public void Register()
                     return surfacePage;
                 }, new HandlerOptions { SelfOnly = true });
 
+            Handle.GET("/WebsiteProvider/surface/{?}/html", (string surfaceId) =>
+            {
+                var html = "<template><template is=\"dom-bind\">\n";
+                var blendingPoints = Db.SQL<WebSection>($"SELECT ws FROM {typeof(WebSection).FullName} ws WHERE ws.Template.Key = ? AND ws.{nameof(WebSection.Deleted)} = ?", surfaceId, false);
```
un-tone: We should not use `Key` property in SQL because of case-insensitive string comparison.

src/WebsiteProvider/Api/ContentHandlers.cs
```diff
@@ -55,6 +57,25 @@ public void Register()
...
+                foreach (WebSection blendingPoint in blendingPoints)
+                {
+                    var name = blendingPoint.Name;
+                    html += $"<starcounter-include slot=\"websiteeditor/{name.ToLowerInvariant()}\" view-model=\"{{{{model.Sections.{name}}}}}\"></starcounter-include>\n";
```
un-tone: I don't like this `name.ToLowerInvariant()`. It is not obvious and not described that slot names should be written in html compositions in lower case, i.e. slot names are case-sensitive.
I think it should be described (in README?) what the requirements we have for slot names and blending points. Maybe we should remove that `name.ToLowerInvariant()` and warn users that blending point names should be written in the same case as slot names. WDYT about it?

src/WebsiteProvider/WebsiteProvider.csproj
```diff
@@ -89,12 +89,7 @@
     <TypedJSON Include="ViewModels\ViewKeeper.json">
       <Generator>MsBuild:Compile</Generator>
     </TypedJSON>
-    <None Include="wwwroot\WebsiteProvider\surfaces\ResponsiveSurface.html" />
-    <None Include="wwwroot\WebsiteProvider\surfaces\SingleBlendingPointSurface.html" />
     <None Include="wwwroot\WebsiteProvider\surfaces\DefaultSurface.html" />
```
un-tone: Why we still have this file? :-) It should be removed, right?

test/WebsiteProvider_AcceptanceHelperOne/WebsiteProvider_AcceptanceHelperOne.csproj
```diff
@@ -99,6 +99,14 @@
...
     <None Include="wwwroot\WebsiteProvider_AcceptanceHelperOne\FormPage.html" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="wwwroot\WebsiteProvider_AcceptanceHelperOne\surfaces\DefaultSurface.html" />
+    <Content Include="wwwroot\WebsiteProvider_AcceptanceHelperOne\surfaces\HolyGrailSurface.html" />
+    <Content Include="wwwroot\WebsiteProvider_AcceptanceHelperOne\surfaces\ResponsiveSurface.html" />
+    <Content Include="wwwroot\WebsiteProvider_AcceptanceHelperOne\surfaces\SidebarSurface.html" />
+    <Content Include="wwwroot\WebsiteProvider_AcceptanceHelperOne\surfaces\SingleBlendingPointSurface.html" />
+    <Content Include="wwwroot\WebsiteProvider_AcceptanceHelperOne\surfaces\TopbarSurface.html" />
+  </ItemGroup>
```
un-tone: DefaultSurface.html, ResponsiveSurface.html and SingleBlendingPointSurface.html are not used in the tests, so should be removed.

un-tone:
Some additional notes:
1. View URI for surfaces should be read-only (or maybe hidden?) on the general page.
2. Tests which cover the new feature should be added. I could take this part if you want.

src/WebsiteEditor/Helpers/DataHelper.cs
PatrykSzwer: You're right, I'll clean this since there will be option in settings which will do same thing but in more controlable way.

src/WebsiteProvider/WebsiteProvider.csproj
PatrykSzwer: Actually I'm not sure about this one. This one was left for backwards compatibility. And now, when I'm thinking about it, maybe we should still have files for all surfaces for same reason? On the other hand I'm not sure if holding to backward compatibility right now is important, since We could just do a release of new version. I would like to know @warpech thoughts about that,

un-tone: Having something only for backward compatibility is applicable only if it is covered with tests or at least some use case. Otherwise, some day this stuff will be inoperable. Just my opinion.

src/WebsiteProvider/Api/ContentHandlers.cs
PatrykSzwer: I've removed `.ToLoverInvariant()`. About that warning for user, do You think that we should add something like `Be aware that blending point names are case sensitive` or just `Warning! blending point names are case sensitive` in blending points view?

PatrykSzwer:
>View URI for surfaces should be read-only (or maybe hidden?) on the general page.

Now when I think about it, I'll just remove this from view, since It's no longer important for end user.

src/WebsiteEditor/Helpers/DataHelper.cs
PatrykSzwer: I liked the idea. It's done.

src/WebsiteProvider/WebsiteProvider.csproj
warpech: These HTML files should be reverted for backwards compatibility for people who have a current configuration that uses the static files.

PatrykSzwer: I've restored those files.

PatrykSzwer:
I've just added fix on really crucial bug here. Before that, when You tried to run websiteEditor without compositionProvider running You would get empty page without anything rendered. That was because of missing slot declarations in "default" declarative-shadow-dom. Right now when user runs website without compositionProvider he will get everything rendered with blending points content separated by the lines. @un-tone please check this change and I'll check yours.

PatrykSzwer:
Your changes/test looks good. Approved.

un-tone:
Really useful fixes :-). Works well. Approved again.

Issue #258, opened 2017-10-26
Apply agreed project structure


warpech:
The project structure does not adhere to our own guidelines presented at https://docs.starcounter.io/cookbook/project-structure/

What I mean:

- folder `src/WebsiteEditor/wwwroot/WebsiteEditor/viewmodels` should be renamed to `views`
- perhaps something else? (please check)

This is low prio.

PatrykSzwer:
Closing as done.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Website/releases/tag/4.1.0

PR #277, opened 2017-11-17
Apply correct folder structure

PatrykSzwer:
for #258

I couldn't find any other mistakes than viewmodels -> views folder name. If You find out that there are some other issues, please let me know.


un-tone:
I found the following:

1. I guess we don't have to have these files https://github.com/Starcounter/Website/tree/master/src/WebsiteEditor/Models

1. Maybe it is not exactly for this GH issue, but this file should be moved https://github.com/Starcounter/Website/blob/master/src/WebsiteEditor/wwwroot/WebsiteEditor/css/holygrail.css because this file is not accessible in the case when WebsiteEditor is not running.

1. https://github.com/Starcounter/Website/blob/master/src/WebsiteEditor/wwwroot/WebsiteEditor/images/cog.svg can be removed since it is not used.

1. The folder `sys` should be removed from the WebsiteEditor project (I mean only from WebsiteEditor.csproj) https://github.com/Starcounter/Website/blob/4543323ba452af1af193cc796e657417d9d54555/src/WebsiteEditor/WebsiteEditor.csproj#L148

1. In the test helper projects there are no "views" directory:
    * https://github.com/Starcounter/Website/tree/master/test/WebsiteProvider_AcceptanceHelperOne/wwwroot/WebsiteProvider_AcceptanceHelperOne
    * https://github.com/Starcounter/Website/tree/master/test/WebsiteProvider_AcceptanceHelperTwo/wwwroot/WebsiteProvider_AcceptanceHelperTwo
Maybe it is not so important but why not.

PatrykSzwer:
> I guess we don't have to have these files /src/WebsiteEditor/Models@master

I have absolutely no idea why those files are there. Why those were even added? Do You know the story behind it?

> Maybe it is not exactly for this GH issue, but this file should be moved /src/WebsiteEditor/wwwroot/WebsiteEditor/css/holygrail.css@master because this file is not accessible in the case when WebsiteEditor is not running.

Let's make this one in another issue. Because we have to add new composition for this etc.

un-tone:
> I have absolutely no idea why those files are there. Why those were even added? Do You know the story behind it?

It seems that these files just left forgotten after moving into Simplified

un-tone:
> Let's make this one in another issue. Because we have to update composition for this etc.

Created #278

PatrykSzwer:
I think I've covered all of your points.

Issue #278, opened 2017-11-20
Move holygrail.css from WebsiteProvider


un-tone:
This file should be moved [WebsiteEditor/wwwroot/WebsiteEditor/css/holygrail.css](https://github.com/Starcounter/Website/blob/master/src/WebsiteEditor/wwwroot/WebsiteEditor/css/holygrail.css) because this file is not accessible in the case when WebsiteEditor is not running.

PatrykSzwer:
Closing as done.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Website/releases/tag/4.1.0

PR #282, opened 2017-11-21
Remove holygrail.css

PatrykSzwer:
For #278

Hey @un-tone could You please take a look here, why with composition in separated file, blending point marks are not generated? I can't find a problem here.


un-tone:
Because you changed slot names, e.g. from `websiteeditor/Header` to `websiteeditor/header`.

PatrykSzwer:
You're right, I forgot to change it after I've copied it from editor. It's fixed now, please check.

Issue #280, opened 2017-11-21
Import and export data should have more informative messages


PatrykSzwer:
Right now, import and export actions have only "Successful" message available. It's clearly not enough and we have to show what actually happens in console, when something went wrong. I'll add information from console output to message box whenever something wrong happen.

PatrykSzwer:
Closing as done.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Website/releases/tag/4.1.0

PR #283, opened 2017-11-22
Error messages for stardump operations

PatrykSzwer:
#280 

I've added error messages, for either errors came from stardump and also ones from starcounter-upload when You try to import non-db file.


src/WebsiteEditor/ViewModels/SettingsPage.json
```diff
@@ -1,7 +1,9 @@
 {
   "Html": "/WebsiteEditor/views/SettingsPage.html",
   "RedirectUrl$": "",
-  "Message": "",
+
+  "Message$": "",
+  "isErrorMessage$": false,
```
un-tone: Why not `IsErrorMessage$`?

src/WebsiteEditor/ViewModels/SettingsPage.json.cs
```diff
@@ -120,10 +124,14 @@ public void ImportDataToDatabase(string filePath)
                             return;
                         }
 
-                        DataHelper.RefreshSurfaceUris();
                         this.ImportInProgress = false;
                         this.ImportButtonText = "Import data";
-                        this.Message = "Import completed!";
+
+                        if (string.IsNullOrEmpty(this.Message))
```
un-tone: I'd recommend to check just `this.isErrorMessage` property here

Issue #287, opened 2017-11-29
Add link to website tutorial video in Readme


PatrykSzwer:
@warpech recorded pretty cool, explanatory video about basically every aspect of Website app. I think there is no cons in adding link to this video into readme so that every new user, who doesn't know about video, will see it right away. 

@warpech WDYT? Are there any arguments against this?

warpech:
That's a very good idea, please do that.

PatrykSzwer:
Done.

PR #288, opened 2017-11-29
Tutorial video in readme

PatrykSzwer:
For #287

I just added video and image with "play arrow". I'm not sure if I should add any sentence before it like "Click here to see website tutorial showing...".
@Mackiovello WDYT?


PatrykSzwer:
I could use an image generated from youtube:
[Website tutorial](http://www.youtube.com/watch?v=dvQHlzoN0wE)

But I didn't because without that arrow I've added, It seems more like picture than a video. WDYT?

PatrykSzwer:
Take a look now.

PR #292, opened 2017-12-08
Changed test for JSONs without Html field

un-tone:
For https://github.com/Starcounter/Home/issues/332.
Related changes: https://github.com/Starcounter/level1/pull/4397


test/WebsiteProvider.Tests/Test/MiddlewareHandlingTests.cs
```diff
@@ -53,11 +55,12 @@ public void MiddlewareHandlesResponse_CookieAdded()
         /// Test if an exception data was rendered when a responded Json doesn't have an Html field.
         /// </summary>
         [Test]
-        public void LoadEmptyJson_ExceptionReturned()
+        public void RequestEmptyJson_ResponseNotWrapped()
         {
             Driver.Navigate().GoToUrl(Config.AcceptanceHelperOneUrl + "/EmptyJson");
 
-            WaitUntil(x => x.PageSource.Contains("System.InvalidOperationException: ScErrInvalidOperation (SCERR1025)"));
+            WaitUntil(x => x.PageSource.Contains("{}"));
```
warpech: Can this be `AreEqual` instead of `Contains`?

ubbeK:
@ubbeK, could TeamCity configuration be changed to automatically build code committed to feature branches?

Yes, @joozek78 has implemented that for `level0` and it will most likely be introduced to `level1`

test/WebsiteProvider.Tests/Test/MiddlewareHandlingTests.cs
un-tone: No, here the PageSource is
```html
<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body><pre style="word-wrap: break-word; white-space: pre-wrap;">{}</pre></body></html>
```

un-tone: Yeah, I should change the test since the PageSource is different for different browwers

un-tone: I removed that checking since it is unnecessary. All we should check is that response is not wrapped.

test/WebsiteProvider.Tests/Test/MiddlewareHandlingTests.cs
```diff
@@ -53,11 +55,10 @@ public void MiddlewareHandlesResponse_CookieAdded()
         /// Test if an exception data was rendered when a responded Json doesn't have an Html field.
         /// </summary>
         [Test]
-        public void LoadEmptyJson_ExceptionReturned()
+        public void RequestEmptyJson_ResponseNotWrapped()
         {
             Driver.Navigate().GoToUrl(Config.AcceptanceHelperOneUrl + "/EmptyJson");
-
-            WaitUntil(x => x.PageSource.Contains("System.InvalidOperationException: ScErrInvalidOperation (SCERR1025)"));
+            Assert.IsFalse(Driver.FindElements(By.ClassName("website-surface")).Any());
```
warpech: This test will keep passing if for some reason:
- we change the class name from `"website-surface"` to anything else.
- the class  `"website-surface"` shows up with delay

I still think that another assertion `WaitUntil(x => x.PageSource.Equals("{}"));` would be good.

un-tone: @warpech fixed

PR #294, opened 2017-12-14
Update paper-collapse-item from main repository

PatrykSzwer:
I've replaced source of paper-collapse-item element from my fork to original repository since they've merged my changes.


warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Website/releases/tag/4.1.0

PR #300, opened 2017-12-27
Fixed wrapping responses

un-tone:
For https://github.com/Starcounter/Website/issues/298


PatrykSzwer:
Changes itself seems good, I've checked it and everything is working correctly with them. :+1:

un-tone:
Finding the test case for this bug turned out harder than fixing it 😞.
I have not done it still.

un-tone:
@warpech Finding the test case for this bug turned out harder than fixing it 😞.
I don't have it ready still.

un-tone:
The fix looks doubtfully. I mean currently I don't understand the reason of the error and why the fix works. Probably it is just a workaround for some Starcounter glitch. I suggest to isolate the issue so we could investigate it deeply.
Actually only https://github.com/Starcounter/Website/pull/300/commits/172981666e99025bce41b39b2569b01a08ead0ab commit relates to the issue so I moved other fixes (https://github.com/Starcounter/Website/pull/300/commits/fcac71d8fe2862086e1b9f8bc41c40651e1cb704) to the separate PR #309.

warpech:
I just discussed with @un-tone that this we should not rush with including this PR. The final version of WebsiteProvider can go without it.

This PR should be closed and properly fixed with tests in BlendingProvider. @un-tone will take care of it next week.

PR #305, opened 2017-12-29
Explain branches

warpech:
This PR explains the current branching model and includes a few other minor fixes.

I'd like to ask everyone in the @Starcounter/webplatformteam to review. If you agree, this will be transferred to KitchenSink and our other apps.


warpech:
Thanks for the review, merging.

Issue #298, opened 2017-12-20
"WebSocket connection closed" after clicking tabs in specific order


PatrykSzwer:
### Steps to reproduce
1. Go to `http://localhost:8080/WebsiteEditor`
2. Click two times at tab `Surface groups`
3. Click on `All routes` tab.
### Expected result
No errors
### Actual result
![image](https://user-images.githubusercontent.com/25042256/34200554-d1490f72-e571-11e7-90f3-0d26e88f5d59.png)

##### Starcounter version 2.3.1.8061

cc: @un-tone

warpech:
@PatrykSzwer please always try to report the visible reason for the error if possible.

 "WebSocket connection closed" is not the reason but the symptom of an exception that happens on the server while applying a patch.

The real reason is `Starcounter.DbException: ScErrTransactionClosed....` and is logged in the full form on the server. The error log is available in Administrator > Logs.

warpech:
PRs #309 and #300 went out of scope of the final release of Website. They should be fixed in BlendingProvider.

PatrykSzwer:
In that case I'll just close this issue, and move it to Blending repository already.

PR #309, opened 2018-01-04
Fixed wrapping

un-tone:
Separated from #300 since it does not really relate to #298 issue.

These lines have typos: https://github.com/Starcounter/Website/blob/70ad491f486fc2ad8c2e4c769f6fb65233fb5187/src/WebsiteProvider/Api/ContentHandlers.cs#L234 and https://github.com/Starcounter/Website/blob/70ad491f486fc2ad8c2e4c769f6fb65233fb5187/src/WebsiteProvider/Api/ContentHandlers.cs#L238

After fixing it one test didn't pass. So the other changes here fixes the case when upper-level wrappers are not being updated even if the original content is.


un-tone:
The fix does not require adding new tests. The existing one already covers it.

PatrykSzwer:
Could You add some small comment/s about those changes? I'm afraid that reason of those will disappear in time.

`. So the other changes here fixes the case when upper-level wrappers are not being updated even if the original content is.`
Which changes Did You mention here? You mean all other changes except replacing `restoredFromSession = true;` with `restoredFromSession = false;` and opposite are for test purposes?

warpech:
I just discussed with @un-tone that this we should not rush with including this PR. The final version of WebsiteProvider can go without it.

This PR should be closed and properly fixed with tests in BlendingProvider. @un-tone will take care of it next week.

Issue #308, opened 2018-01-03
Remove viewport meta tag creation script from ResponsiveSurface.html


alshakero:
It's redundant now.

Lines of interest: https://github.com/Starcounter/Website/blob/70ad491f486fc2ad8c2e4c769f6fb65233fb5187/src/WebsiteProvider/wwwroot/WebsiteProvider/surfaces/ResponsiveSurface.html#L35-L42

More on this in this commit https://github.com/Starcounter/level1/commit/61e4eaaaa6607c07389505f77ab34ce1b8ed6849.

PatrykSzwer:
This issue is moved to https://github.com/Starcounter/Blending/issues/7

PR #310, opened 2018-01-04
Removed redundant viewport meta tag creation script

PatrykSzwer:
For #308


alshakero:
We only added this meta tag to 2.4. This PR base should be `master-2.4`, also, same issue applies to Blending app.

PatrykSzwer:
You're right, my bad. I've rebased this PR on `master-2.4`

PatrykSzwer:
Since I've released final version of website already, and this change isn't crucial, let's move this to blending app already.

